PR #3, opened 2015-09-01
adding hand wrist joints to arm controllers

kirstyellis:


ugocupcic:
Needs the same changes for left arm

PR #13, opened 2015-09-11
Fixing shippable tests

ugocupcic:


ugocupcic:
let's see if this fixes shippable.

PR #19, opened 2015-09-16
added (almost) empty function to return a couple of named poses.

dg-shadow:


sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -129,6 +129,15 @@ def plan_to_named_target(self, name):
         self._move_group_commander.set_named_target(name)
         self.__plan = self._move_group_commander.plan()
 
+    def get_all_named_targets(self):
```
ugocupcic: what about simply `get_named_targets`?

PR #21, opened 2015-09-16
Fixed the prefix for the rh

beatrizleon:
Now this works:
roslaunch sr_moveit_hand_config demo.launch 

But it will need to be changed to another hand. This will be solved with the more general solution mentioned in: https://github.com/shadow-robot/sr-ros-interface/issues/215


VahidAminZ:
will there be a lh at close future?

PR #37, opened 2015-09-22
F add hand kinematics example

beatrizleon:
Added the waypoint function to the robot commander.
Also modified the init of hand and robot commander to allow to start a group for controlling the fingers.
Added an example to show how to plan a trajectory for the first finger following waypoints (more or less a square) showed in the following video:

gifrecord_2015-09-22_162325

It is usually ok, but sometimes it fails to plan for the inital open finger posture. Also, when executing the plans, they are executed but I am getting error messages of aborted plans because a velocity tolerance error.


sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
```diff
@@ -0,0 +1,73 @@
...
+
+pose = geometry_msgs.msg.Pose()
+pose.position.x = 0.057038
+pose.position.y = -0.13281
+pose.position.z = 0.37172
+pose.orientation.x = 0.60863
+pose.orientation.y = 0.05909
+pose.orientation.z = -0.044238
+pose.orientation.w = 0.79001
+waypoints.append(copy.deepcopy(pose))
```
dg-shadow: Does this need to be deepcopy? Surely a new Pose is being created each time... Not sure if this is a problem...

beatrizleon: Fixed

sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
```diff
@@ -0,0 +1,73 @@
...
+# Open fingers
+group_id = "fingers"
+hand_commander = SrHandCommander(name=group_id)
+hand_commander.move_to_named_target("fingers_open")
+
+group_id = "first_finger"
+hand_commander = SrHandCommander(name=group_id)
+
+rospy.sleep(2)
+print("\nMoving Index finger following a set of waypoints")
```
dg-shadow: Maybe should be a rospy.loginfo rather than print?

PR #43, opened 2015-09-25
we want to be able to change those mappings easily if we were to have…

ugocupcic:
… a bimanual system with lite hands or what not


ugocupcic:
goes with https://github.com/shadow-robot/sr-config/pull/53

PR #85, opened 2015-11-04
F pose planning

dg-shadow:
Changes needed for sr_manipulation


sr_robot_commander/src/sr_robot_commander/sr_arm_commander.py
```diff
@@ -95,3 +97,23 @@ def _set_ground(self):
...
+            return None
+
+        start_pose = self._move_group_commander.get_current_pose().pose
+
+        (self._SrRobotCommander__plan, fraction) = self._move_group_commander.compute_cartesian_path(
+            [start_pose, target_pose], eef_step, jump_threshold)
+
+        if fraction < min_fraction:
+            rospy.logerr("Couldn't reach enough waypoints, only %f" % fraction)
+            self._SrRobotCommande__path = None
```
AndriyPt: @dg-shadow are you playing with name mangling?

dg-shadow: Is that a trick question?

dg-shadow: Also, yes :/

dg-shadow: This function is only useful for arm, but needs access to __plan from RobotCommander... Guess the clean thing to do is add a protected function in SrRobotCommander then access it from SrArmCommander... How much do you care?

AndriyPt: well, you might solve it in more OOP way e.g. by "protected" setter function (started by single underscore by convention of some group of people)

AndriyPt: @dg-shadow it would be nice to have such protected function. Even something like _clean_plan(). Name mangling algorithm might change in the future versions of Python and your code will not work :)

PR #105, opened 2015-11-20
A few small fixes

toliver:
I can be merged.

Not closing the issue 95 yet though. As I need to check some other examples.


sr_moveit_hand_config/config/ompl_planning_template.yaml
```diff
@@ -183,6 +183,7 @@ fingers:
     - TRRTkConfigDefault
     - PRMkConfigDefault
     - PRMstarkConfigDefault
+  projection_evaluator: joints(FFJ4,FFJ3)
```
ugocupcic, line 186: not needed for the other fingers?

toliver, line 186: The planner was failing, complaining that no `projection_evaluator` was defined. So I added one and it worked.

Not sure if it makes sense to add the 2 first joints of every finger. I would wait for that to become necessary. Right now the `generate_moveit_config.py` script wouldn't deal well with a config the includes more than 2 joints.

PR #113, opened 2015-11-26
F#87 store pose example hand lite

### LINKED ISSUE #87: Store position script for hand lite
The generic one doesn't seem to work.

beatrizleon:
Fixes #87 

This example now is able to store all hands:
rosrun sr_example sr_store_hand.py

Other similar test were removed.

Pack position is different for old biotac hand.

Tested with all hands in simulation and with real extra lite.

Added test to check that "biotac_decimated.dae" exists for biotac hands


sr_moveit_hand_config/config/shadowhands_prefix.srdf.xacro
```diff
@@ -227,20 +250,39 @@
...
+    <xacro:if value="$(arg rf)">
+        <xacro:finger_state_pack  prefix="$(arg prefix)" joint_prefix="RF" lf="0" is_biotac="$(arg is_biotac)"/>
+    </xacro:if>
+    <xacro:unless value="$(arg is_lite)">
+      <xacro:if value="$(arg lf)">
+          <xacro:finger_state_pack  prefix="$(arg prefix)" joint_prefix="LF" lf="1" is_biotac="$(arg is_biotac)"/>
+      </xacro:if>
+    </xacro:unless>
+  </group_state>
+  <group_state name="fingers_pack_thumb_open" group="$(arg prefix)fingers">
```
toliver, line 270: This state name seems to be defined twice in this file.

beatrizleon, line 270: Yes, once for the hand group and once for the fingers group, so you can use it in either

PR #114, opened 2015-11-30
F#103 set planning frame

### LINKED ISSUE #102: Increase start_state_max_bounds
to avoid the following warning, that prevents moveit from finding a plan:

``` bash
[ WARN] [1448035168.814770579]: Joint 'rh_RFJ2' from the starting state is outside bounds by a significant margin: [ 1.70883 ] should be in the range [ 0 ], [ 1.5708 ] but the error above the ~start_state_max_bounds_error parameter (currently set to 0.1)
```

The reason this happens is that xFJ1 and XFJ2 are not physically constrained between 0 and 90 deg. And the position controller for xFJ0 actually controls the addition of both positions, but not each of them independently, whereas moveit checks their limits independently.

### LINKED ISSUE #103: Add a way to choose planning frame to RobotCommander
Currently `plan_to_waypoints_target` takes poses, and not StampedPoses (due to the underlying `compute_cartesian_path`).

We should allow the user to specify what is the planning frame, so that the Poses can be used in a sensible way.

Otherwise (like it happens in the `sr-right_hand_follow_finger_trajectory` example) the poses are dependant on the robot URDF. E.g. the pose of the fingertip is relative to the base of the robot, but the base (world) will probably the forearm base if it's a hand, or the arm base if it's a hand+arm. Also, the pose of the finger tip w.r.t. the base of the hand is different for a hand E than for a hand lite.

The easiest fix seems to be to allow the user to set the desired frame from the script before planning.

beatrizleon:
Merge after https://github.com/shadow-robot/sr_interface/pull/113

Fixes #103 
Fixes #102

### DIFF
--- a/sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
+++ b/sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
@@ -15,60 +15,74 @@ hand_parameters = hand_finder.get_hand_parameters()
 prefix = hand_parameters.mapping.values()[0]
 hand_serial = hand_parameters.mapping.keys()[0]
 
-# Open fingers
-group_id = prefix + "_fingers"
-hand_commander = SrHandCommander(name=group_id)
-hand_commander.move_to_named_target("fingers_open")
+# Close all fingers except thumb
+hand_commander = SrHandCommander()
+hand_commander.move_to_named_target("fingers_pack_thumb_open")
 
+# Change to first finger group
 group_id = prefix + "_first_finger"
 hand_commander = SrHandCommander(name=group_id)
 
-rospy.sleep(2)
+# Open first finger
+hand_commander.move_to_named_target("first_finger_open")
+
+rospy.sleep(4)
+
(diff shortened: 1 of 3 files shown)


sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
```diff
@@ -15,60 +15,74 @@
...
 rospy.loginfo("\nMoving Index finger following a set of waypoints")
 
 waypoints = []
 
+# Move horizontally
+start_pose = geometry_msgs.msg.Pose()
+start_pose.position.x = 0.034
+start_pose.position.y = -0.002
+start_pose.position.z = 0.197
+start_pose.orientation.w = 1.00
```
toliver, line 40: I see that the orientation is the same for all the waypoints. Does that mean that moveit will ignore the orientation if it's not achievable?

toliver, line 40: As discussed with @beatrizleon  the orientation is ignored by the IK for the fingers, as most orientations would be infeasible anyway. This will be documented.

PR #115, opened 2015-12-01
F#104 change name current pose

### LINKED ISSUE #104: get_current_pose method of robot_commander is badly named
It should be called something like "get_movegroup_state"

Changing it will involve grepping a at least all of sr_interface to find places where the old style method is used. It's a relatively recent function so it's probably not used outside sr_interface anywhere yet.

beatrizleon:
Fixes #104 

Renamed the current get_current_pose to get_current_state and replaced where it was used.

Created a new get_current_pose to actually get the pose of the end effector. It takes the reference frame as an argument. If it is given, it uses tf to find the pose if not it uses the default get_current_pose from moveit which uses the planning_frame as reference (/world)

### DIFF
--- a/sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
+++ b/sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
@@ -32,14 +32,10 @@ rospy.loginfo("\nMoving Index finger following a set of waypoints")
 
 waypoints = []
 
-# Move horizontally
-start_pose = geometry_msgs.msg.Pose()
-start_pose.position.x = 0.034
-start_pose.position.y = -0.002
-start_pose.position.z = 0.197
-start_pose.orientation.w = 1.00
+start_pose = hand_commander.get_current_pose(reference_frame=prefix + "_palm")
 waypoints.append(start_pose)
 
+# Move horizontally
 pose = geometry_msgs.msg.Pose()
 pose.position.x = 0.067
 pose.position.y = -0.010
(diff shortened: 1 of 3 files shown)


sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -213,13 +216,38 @@ def get_named_target_joint_values(self, name):
 
         return output
 
-    def get_current_pose(self):
+    def get_current_pose(self, reference_frame=None):
```
ugocupcic, line 219: add a docstring to both functions to clarify - especially the @return type

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -213,13 +216,38 @@ def get_named_target_joint_values(self, name):
...
+                current_pose.orientation.w = rot[3]
+                return current_pose
+            except (tf.Exception, tf.LookupException, tf.ConnectivityException, tf.ExtrapolationException):
+                rospy.logwarn("Couldn't get the pose from " + self._move_group_commander.get_end_effector_link() +
+                              " in " + reference_frame + " reference frame")
+            return None
+        else:
+            return self._move_group_commander.get_current_pose().pose
+
+    def get_current_state(self):
```
ugocupcic, line 250: here too

sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
```diff
@@ -59,26 +55,30 @@
...
+waypoints.append(pose2)
+
+pose3 = geometry_msgs.msg.Pose()
+pose3.position.x = 0.033
+pose3.position.y = -0.068
+pose3.position.z = 0.151
+pose3.orientation.w = 1.00
+waypoints.append(pose3)
+
+waypoints.append(pose2)
```
toliver, line 79: Why are we appending them in reverse order?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
toliver, line 219: Could you add documentation to the method?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -213,13 +216,38 @@ def get_named_target_joint_values(self, name):
...
         joint_names = self._move_group_commander._g.get_active_joints()
         joint_values = self._move_group_commander._g.get_current_joint_values()
 
         return dict(zip(joint_names, joint_values))
 
-    def get_current_pose_bounded(self):
+    def get_current_state_bounded(self):
```
toliver, line 260: Doc

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -213,13 +216,38 @@ def get_named_target_joint_values(self, name):
 
         return output
 
-    def get_current_pose(self):
+    def get_current_pose(self, reference_frame=None):
+        current_pose = geometry_msgs.msg.Pose()
```
toliver: This assignation can be inside the `try`.

sr_example/scripts/sr_example/hand_examples/sr_right_hand_follow_finger_trajectory.py
beatrizleon, line 79: I created a trajectory first moving horizontally (abduction/adduction) of the finger and then moving vertically (flexing the finger) so to do the latter I appended three poses and then, to move back to the initial pose, I appended them but in reverse order.

toliver, line 79: Sorry I didn't see the first 3 appends. :)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -213,13 +216,52 @@ def get_named_target_joint_values(self, name):
...
+        """
         joint_names = self._move_group_commander._g.get_active_joints()
         joint_values = self._move_group_commander._g.get_current_joint_values()
 
         return dict(zip(joint_names, joint_values))
 
-    def get_current_pose_bounded(self):
+    def get_current_state_bounded(self):
+        """
+        Get the current joint state of the group being used, checking that they are within each joint limits.
```
toliver: A bit more accurate to say `enforcing` rather than checking.

PR #126, opened 2015-12-21
Fixes #96 - interface tests

### LINKED ISSUE #96: Write proper tests for the commander classes
At least test all the methods.

Preferably test all methods with a representative set of parameters. Some configuration changes that break things only show when using certain joint groups, or other combinations of parameters and circumstances.

We could try to maintain a reasonable set of tests, but it would also be useful to be able to run the examples in sr_examples as tests (not sure how to do that though).

ugocupcic:
Not working well since they all segfault (see #51). Not sure if we should merge this.


beatrizleon:
Closing as a new pr is opened rebased for kinetic

PR #129, opened 2016-01-25
add world parameter to some launch files

kirstyellis:
please can I add a 'world' argument to these launch files? I think it will be super useful for G and Beatriz's project of creating cool simulations, where I would imagine we will want to switch worlds!

### DIFF
--- a/sr_robot_launch/launch/sr_ur10arm_hand.launch
+++ b/sr_robot_launch/launch/sr_ur10arm_hand.launch
@@ -47,10 +47,9 @@
 
   <!-- SIMULATED ROBOTS -->
   <group if="$(arg sim)">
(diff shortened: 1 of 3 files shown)


sr_robot_launch/launch/sr_ur10arm_hand.launch
```diff
@@ -46,7 +47,7 @@
 
   <!-- SIMULATED ROBOTS -->
   <group if="$(arg sim)">
-    <arg name="world_name" value="$(find sr_description)/other/worlds/shadowhand.world"/>
+    <arg name="world_name" value="$(arg world)"/>
```
toliver: The arg `world_name` is now redundant.

PR #150, opened 2016-04-12
F remove deprecated moveit cfgs

kirstyellis:


sr_interface/package.xml
```diff
@@ -19,8 +19,6 @@
   <run_depend>sr_robot_commander</run_depend>
   <run_depend>sr_example</run_depend>
   <run_depend>bimanual_moveit_config</run_depend>
```
beatrizleon: bimanual sistem can be deleted as well

PR #165, opened 2016-06-02
adding stomp and sbpl planning pipelines

ugocupcic:


ugocupcic:
SBPL and STOMP dependencies are left out of the repository.rosinstall by design as discussed with @toliver.

PR #168, opened 2016-06-06
Fixes #64 - planners benchmarking

ugocupcic:
This will still need to be improved: 
- checking planners reliability
- ordering the final table


toliver:
This can be merged isn't it @ugocupcic ?

ugocupcic:
Yes.

PR #190, opened 2016-10-12
proposed changed to add initial position to xacro

beatrizleon:
In order to change the initial position of the robot in gazebo and moveit, it needs to be part of the robot urdf. Here is a proposed change to include it to the xacro as an optional parameter that can be passed through the launch file.

I think if you agree, we could also add an optional x and y

### DIFF
--- a/sr_multi_description/urdf/right_srhand_ur10_joint_limited.urdf.xacro
+++ b/sr_multi_description/urdf/right_srhand_ur10_joint_limited.urdf.xacro
@@ -8,7 +8,12 @@
     <joint name="world_joint" type="fixed">
         <parent link="world"/>
         <child link="ra_base_link"/>
-        <origin xyz="0.0 0.0 0.0" rpy="0.0 0.0 0.0"/>
+        <xacro:if value="$(arg initial_z)">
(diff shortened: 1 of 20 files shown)


sr_multi_description/urdf/right_srhand_ur10_joint_limited.urdf.xacro
```diff
@@ -8,7 +8,12 @@
     <joint name="world_joint" type="fixed">
         <parent link="world"/>
         <child link="ra_base_link"/>
-        <origin xyz="0.0 0.0 0.0" rpy="0.0 0.0 0.0"/>
+        <xacro:if value="$(arg initial_z)">
+        	<origin xyz="0.0 0.0 $(arg initial_z)" rpy="0.0 0.0 0.0"/>
+		</xacro:if>
+        <xacro:if value="$(arg initial_z)">
+        	<origin xyz="0.0 0.0 0.0" rpy="0.0 0.0 0.0"/>
+		</xacro:if>
```
toliver: The indentation is not correct.

Also, it seems that this modification should also be applied to other urdf files (like the one for ur5)

PR #217, opened 2016-11-30
Update repository.rosinstall

ethanfowler:
The following package references have been updated, as the packages have to be modified for the kinetic migration:
- sr_core
- sr_interface
- sr-ros-interface-ethercat

Removed references to now-outdated indigo-devel moveit_<package> repos. These repos have been consolidated (see https://github.com/davetcoleman/moveit_merge/blob/master/README.md). The package names have not changed, so package dependencies should still resolve.

Removed reference to ros_controllers branch F#182_partial_trajectory_mod. This fix is already merged in ros_controllers released kinetic.

### DIFF
--- a/repository.rosinstall
+++ b/repository.rosinstall
@@ -6,7 +6,7 @@
 - git:
     local-name: sr_core
     uri: https://github.com/shadow-robot/sr_core
-    version: 'indigo-devel'
+    version: 'kinetic-devel'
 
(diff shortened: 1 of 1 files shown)


beatrizleon:
But, we still need a reference to the consolidated moveit repo, right?

beatrizleon:
Replaced by #226

PR #221, opened 2016-12-29
Added funtions to move to a pose unsafe getting ik

beatrizleon:
Added the ```get_ik``` and ```move_to_pose_value_target_unsafe```  functions. Also changed the name of some functions to indicate that they can be publicly used.


toliver (approved):
Cool!

PR #223, opened 2016-12-30
depend on packages instead of metapackages

### LINKED ISSUE #222: fix trak_ik warning
```
WARNING: package "sr_multi_moveit_config" should not depend on metapackage "trac_ik" but on its packages instead
WARNING: package "sr_moveit_hand_config" should not depend on metapackage "trac_ik" but on its packages instead
```

beatrizleon:
Fixes #222


beatrizleon:
Replaced by #228

PR #226, opened 2017-01-13
Making sr_interface build 

beatrizleon:


repository.rosinstall
```diff
@@ -21,20 +21,19 @@
 - git:
     local-name: universal_robot
     uri: 'https://github.com/ros-industrial/universal_robot.git'
-    version: kinetic-devel
+    version: 'kinetic-devel'
```
toliver: I think that our default format is without single quote.
We should remove it from the other lines instead.

repository.rosinstall
```diff
@@ -31,10 +31,14 @@
 - git:
     local-name: lma_kinematics_plugin
     uri: 'https://github.com/shadow-robot/lma_kinematics_plugin'
```
toliver: Can we test it without this source dependency?
I think this package is released for kinetic.
As far as i remember the changes we made in this ik_with_joint _limits branch are not very important, so it's worth testing the original repo.

repository.rosinstall
```diff
@@ -31,10 +31,14 @@
...
-    version: F_ik_with_joint_limits
+    version: F_fixing_kinetic_build
+
+- git:
+    local-name: trac_ik
```
toliver, line 37: Is this not released then?

beatrizleon: Not yet in kinetic. This is what I get ```sudo apt-cache search trac-ik``` in indigo:
```ros-indigo-trac-ik - The ROS packages in this repository were created to provide an improved alternative Inverse Kinematics solver to the popular inverse Jacobian methods in KDL.
ros-indigo-trac-ik-examples - This package contains the source code for testing and comparing trac_ik
ros-indigo-trac-ik-kinematics-plugin - A MoveIt! Kinematics plugin using TRAC-IK
ros-indigo-trac-ik-lib - TRAC-IK is a faster, significantly more reliable drop-in replacement for KDL's pseudoinverse Jacobian solver.
ros-jade-trac-ik - The ROS packages in this repository were created to provide an improved alternative Inverse Kinematics solver to the popular inverse Jacobian methods in KDL.
ros-jade-trac-ik-examples - This package contains the source code for testing and comparing trac_ik
ros-jade-trac-ik-kinematics-plugin - A MoveIt! Kinematics plugin using TRAC-IK
ros-jade-trac-ik-lib - TRAC-IK is a faster, significantly more reliable drop-in replacement for KDL's pseudoinverse Jacobian solver.
```
But nothing on kinetic

toliver: But kinetic packages only appear in xenial and later Ubuntu. If you chek on Trusty you won't find them.
It seems that trac-ik is released on kinetic see https://github.com/ros/rosdistro/blob/master/kinetic/distribution.yaml#L6280

repository.rosinstall
toliver: Actually it seems that now this is part of the moveit repository.
See https://github.com/ros-planning/moveit/tree/kinetic-devel/moveit_kinematics/lma_kinematics_plugin

Can we test how it works without this source dependency (using the lma plugin in moveit).

beatrizleon: New issue created to test this in a separate PR: #231

repository.rosinstall
beatrizleon: Issue created to remove this after the next ROS package sync #232

PR #230, opened 2017-01-25
Generate robot srdf with jade xacro on kinetic

beatrizleon:
Merge after #226 
Same as #185 but against kinetic-devel

### DIFF
--- a/repository.rosinstall
+++ b/repository.rosinstall
@@ -33,6 +33,7 @@
     uri: https://github.com/shadow-robot/lma_kinematics_plugin
     version: F_fixing_kinetic_build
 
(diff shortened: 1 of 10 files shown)


repository.rosinstall
```diff
@@ -34,6 +34,11 @@
     version: F_fixing_kinetic_build
 
 - git:
```
toliver: Please add comment that this dependency will not be needed when version x.y.x of ros_controllers is released.

toliver (changes requested):

toliver:
Only that small comment. We're getting closer!!

PR #237, opened 2017-02-03
Fixes mongodb kinetic

### LINKED ISSUE #233: Mongodb interface was changed in kinetic
There is a mongodb plugin here:
https://github.com/ros-planning/warehouse_ros_mongo

However we need to follow the steps detailed in the Readme to build it: https://github.com/ros-planning/warehouse_ros_mongo/blob/jade-devel/README.md

There are plans to replace mongodb for another database but nothing ready yet:
https://github.com/ros-planning/moveit/issues/123

I am not sure if we should wait until something else is realease?

beatrizleon:
merge after #230 

Solves  #233, still specific build instructions are needed:
https://github.com/ros-planning/warehouse_ros_mongo/blob/jade-devel/README.md

### DIFF
--- a/repository.rosinstall
+++ b/repository.rosinstall
@@ -28,11 +28,6 @@
     uri: https://github.com/shadow-robot/sr_ur_arm.git
     version: kinetic-devel
 
-- git:
-    local-name: lma_kinematics_plugin
-    uri: https://github.com/shadow-robot/lma_kinematics_plugin
-    version: F_fixing_kinetic_build
-
 # Only needed for version 0.11.2. It can be removed for higher versions which will allow the use of partial trajectories
 - git:
     local-name: ros_controllers
--- a/repository.rosinstall
+++ b/repository.rosinstall
@@ -47,3 +47,13 @@
     local-name: shadow_robot_ethercat
     uri: https://github.com/shadow-robot/sr-ros-interface-ethercat.git
     version: kinetic-devel
+
+- git:
+    local-name: warehouse_ros_mongo
+    uri: https://github.com/ros-planning/warehouse_ros_mongo.git
+    version: jade-devel
+
+- git:
+    local-name: mongo-cxx-driver
+    uri: https://github.com/mongodb/mongo-cxx-driver.git
+    version: 26compat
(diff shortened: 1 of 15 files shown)


toliver:
So let's wait until jenkins build passes to merge this one.

repository.rosinstall
```diff
@@ -48,3 +48,18 @@
...
+    local-name: warehouse_ros_mongo
+    uri: https://github.com/ros-planning/warehouse_ros_mongo.git
+    version: jade-devel
+
+- git:
+    local-name: warehouse_ros
+    uri: https://github.com/ros-planning/warehouse_ros.git
+    version: jade-devel
+
+- git:
```
toliver: This is not a ROS package, so it should be installed through some other mechanism. (e.g. ansible role)

toliver (changes requested):

repository.rosinstall
beatrizleon: ok, deleted

sr_moveit_hand_config/package.xml
```diff
@@ -32,8 +32,9 @@
   <run_depend>trac_ik_kinematics_plugin</run_depend>
   <build_depend>trac_ik_lib</build_depend>
   <run_depend>trac_ik_lib</run_depend>
-
-
+  <build_depend>mongodb</build_depend>
+  <run_depend>mongodb</run_depend>
+  <build_depend>scons</build_depend>
```
toliver: Do this package really depend on scons, or is this a trick to facilitate the installation of mongodb?
If so, please add a comment in the package.xml

toliver (changes requested):
If it partially solves the issue, can you comment on what is missing?

sr_moveit_hand_config/package.xml
beatrizleon: Yes, it is only needed for installing the mongo-cxx-driver. As we are not longer installing it here, I am going to remove the dependency

beatrizleon:
I updated pr description and add a link to instructions to build it

PR #240, opened 2017-02-06
F save targets

dg-shadow:


sr_robot_commander/src/sr_robot_commander/sr_robot_state_saver.py
```diff
@@ -10,22 +10,45 @@
...
 
         self.__name = name
-
         self.__hand_h = hand_h
-
-        if hand_or_arm == "arm":
-            self.__commander = SrArmCommander()
-
-        elif hand_or_arm == 'hand':
+        self.__save_target = save_target
```
toliver: I hate these double leading underscores. We should always use single underscores.

dg-shadow: they have a purpose, but ok ;p

dg-shadow: As in, double underscore is technically more "correct" but happy to use the dirty, easier to type, single underscore if you insist

dg-shadow: Although either way, changing to all single _ would make the diff quite messy. maybe merge this one and I'll do a separate pr to change the underscores?

sr_example/scripts/sr_example/hand_examples/sr_hand_optoforce_tactile_example.py
```diff
@@ -1,24 +0,0 @@
-#!/usr/bin/env python
```
toliver: Wow. Why does this appear as deleted?

toliver (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_robot_state_saver.py
toliver: Yes, indeed that would be a separate PR. And I mean it mainly for new files.

About it being more correct. that's far from true.

sr_example/scripts/sr_example/hand_examples/sr_hand_optoforce_tactile_example.py
dg-shadow: No idea o.O

beatrizleon:
Closing this pr as #257 is now opened for kinetic

PR #241, opened 2017-02-07
How was this even working at all??

dg-shadow:
Arm commander methods were calling themselves instead of parent

### DIFF
--- a/sr_robot_commander/src/sr_robot_commander/sr_arm_commander.py
+++ b/sr_robot_commander/src/sr_robot_commander/sr_arm_commander.py
@@ -53,7 +53,7 @@ class SrArmCommander(SrRobotCommander):
         @param end_effector_link - name of the end effector link
         @param wait - should method wait for movement end or not
         """
-        self.move_to_position_target(xyz, end_effector_link, wait=wait)
+        super(SrArmCommander, self).move_to_position_target(xyz, end_effector_link, wait=wait)
 
     def plan_to_position_target(self, xyz, end_effector_link=""):
         """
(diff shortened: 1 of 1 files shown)


sr_robot_commander/src/sr_robot_commander/sr_arm_commander.py
```diff
@@ -53,7 +53,7 @@ def move_to_position_target(self, xyz, end_effector_link="", wait=True):
         @param end_effector_link - name of the end effector link
         @param wait - should method wait for movement end or not
         """
-        self.move_to_position_target(xyz, end_effector_link, wait=wait)
+        super(SrArmCommander, self).move_to_position_target(xyz, end_effector_link, wait=wait)
```
toliver: Actually, why is it defined at all in SrArmCommander?

toliver (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_arm_commander.py
dg-shadow: Not sure...

dg-shadow: Could just remove I guess...

toliver: Probably these methods were copied to RobotCommander at some point, but instead of deleting them from here they were modified to look like this.

I think we should delete them from here.

dg-shadow (commented):
Changed as requested

PR #245, opened 2017-03-02
Improved end effectors and added function to save urdf and srdf to specific location

beatrizleon:


beatrizleon:
Replaced by #251

PR #257, opened 2017-05-08
F kinetic save targets

dg-shadow:


sr_robot_commander/launch/grasp_saver_unsafe.launch
```diff
@@ -0,0 +1,12 @@
+<launch>
+  <arg name="name"/>
+  <arg name="hand_h" default="false"/>
+  <arg name="which" default="all"/>
+  <arg name="save_target" default="false"/>
```
toliver, line 5: Should we default to True?
That seems to be what we want most of the times?
Otherwise we should make very good publicity of this parameter. To create demos this is the configuration we should use.

toliver (changes requested):

sr_robot_commander/launch/grasp_saver_unsafe.launch
dg-shadow: Yeah, maybe. We should also add a switch to the gui (the usual way for most people to use this). I think I'm the only person who ever uses that launch file tbh.

On the other hand, this way means you can't save from teach mode (no target in teach mode) so not sure

Can we just merge cos I need this feature and have another story to add it to the gui?

toliver (approved):

PR #258, opened 2017-05-08
F read controllers from yaml

giusebar:
DO NOT MERGE


sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -579,40 +563,46 @@ def move_to_joint_value_target_unsafe(self, joint_states, time=0.002,
...
     def action_is_running(self):
         return self._action_running
 
     def _action_done_cb(self, terminal_state, result):
         self._action_running = False
 
-    def _call_action(self, goal):
-        self._set_up_action_client()
+    def _call_action(self, goals):
+        self._set_up_action_client(self._controllers)
```
toliver: There is no need to set up the client again, as it was already set up in the constructor.

It seems that this bug is currently in kinetic-devel.

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -579,40 +563,46 @@ def move_to_joint_value_target_unsafe(self, joint_states, time=0.002,
...
         self._action_running = True
-        self._client.send_goal(goal, self._action_done_cb)
+
+        for i, client in enumerate(self._clients.keys()):
+            self._clients[client].send_goal(goals[i], self._action_done_cb)
```
toliver: Using the same callback function for both goals can be an issue. When the first goal finishes we will have `self._action_running = True` which will be misleading if the second (or nth) action hasn't finished.

toliver:
There is an extra file that shouldn't be there /#sr_hand_commander.py#

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -648,15 +652,34 @@ def run_joint_trajectory_unsafe(self, joint_trajectory, wait=True):
...
+
+            for i, joint in enumerate(joint_trajectory.joint_names):
+                if joint in controller_joints:
+                    indices_of_joints_in_this_controller.append(i)
+
+            goal.trajectory.joint_names = [
+                joint_trajectory.joint_names[i] for i in indices_of_joints_in_this_controller]
+
+            for point in goal.trajectory.points:
+                point.positions = [point.positions[i] for i in indices_of_joints_in_this_controller]
```
toliver, line 671: We should also copy point.velocities and point.efforts if present.

toliver (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
dg-shadow: Good catch. Done

toliver (approved):

PR #68, opened 2015-10-21
F reverse trajectory

dg-shadow:
Simple routine to reverse previous trajectory. Simply planes to state defined by start of trajectory given

PR #259, opened 2017-05-09
Change urdf_parser_py import to upstream version

Jntzko:
This relates to the issue [change hand_finder back to use urdfdom instead of local copy after new release of urdfdom #19](https://github.com/shadow-robot/sr_core/issues/19) in the sr_core repo.

The local version of the urdf parser was added because the upstream version
didn't handle hardwareInterfaces in the transmission joints and actuators.

Since the changes of [fixed transmission parser to match specification #68](https://github.com/ros/urdfdom/pull/68), the upstream urdf parser handles TransmissionJoints with optional hardwareInterface elements.

Hardware interfaces in the actuator element shouldn't be used anymore and
are not used in the shadow repos.

### DIFF
--- a/sr_moveit_hand_config/scripts/sr_moveit_hand_config/generate_hand_srdf.py
+++ b/sr_moveit_hand_config/scripts/sr_moveit_hand_config/generate_hand_srdf.py
@@ -45,7 +45,7 @@ import rospkg
 from xacro import set_substitution_args_context
 from rosgraph.names import load_mappings
 
-from sr_utilities.local_urdf_parser_py import URDF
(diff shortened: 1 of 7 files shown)


beatrizleon:
Hi @Jntzko Thanks for the pr. We wanted to ask you if you really need these changes in indigo or would it be ok if we do the changes in kinetic?

Jntzko:
We are working with the shadow hand integrated on a PR2, which runs with indigo. An update to kinetic is not planned, so it would be great to have this changes also in indigo. :)

beatrizleon:
Closing this as we rebased for latest indigo with #298

PR #264, opened 2017-06-06
Changes for SAT

ethanfowler:
The changes to the robot_launch parameters are simple and won't affect anything else (the default value of robot_state_publisher's use_tf_static is maintained).

However I understand that the increase in trajectory_execution/allowed_start_tolerance is contentious. I can't see another way to get grasping to work on sr_interface:kinetic-devel though. @toliver, @beatrizleon, thoughts?

### DIFF
--- a/sr_multi_moveit/sr_multi_moveit_config/launch/trajectory_execution.launch.xml
+++ b/sr_multi_moveit/sr_multi_moveit_config/launch/trajectory_execution.launch.xml
@@ -13,7 +13,7 @@
   <!-- Allow more than the expected execution time before triggering a trajectory cancel (applied after scaling) -->
   <param name="trajectory_execution/allowed_goal_duration_margin" value="0.5"/> <!-- default 0.5 -->
   <!-- Allowed joint-value tolerance for validation that trajectory's first point matches current robot state -->
-  <param name="trajectory_execution/allowed_start_tolerance" value="0.03"/> <!-- default 0.01 -->
+  <param name="trajectory_execution/allowed_start_tolerance" value="6.0"/> <!-- default 0.01 -->
   
   <!-- Load the robot specific controller manager; this sets the moveit_controller_manager ROS parameter -->
   <arg name="moveit_controller_manager" default="shadowhand_motor" />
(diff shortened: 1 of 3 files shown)


sr_robot_launch/launch/sr_right_ur10arm_hand.launch
```diff
@@ -27,6 +27,10 @@
   <arg name="robot_description" default="$(find sr_multi_description)/urdf/right_srhand_ur10_joint_limited.urdf.xacro"/>
   <arg name="robot_config_file" default="$(find sr_multi_moveit_config)/config/robot_configs/right_sh_ur10.yaml"/>
 
+  <!-- When true, robot_state_publisher publishes static frames to /tf_static. This can cause problems with the current rospy tf API,
+       which can't see /tf_static due to a bug (that will be fixed by https://github.com/ros/geometry/pull/134 once merged). -->
```
toliver, line 31: Sadly this is unnecessary. 
The source of the problem is that you are trying to use tf instead of using tf2 (the new library that suppors /tf_static).
Again this is probably the very common issue of obsolete documentation in important pages of the ROS wiki.
That is, landing [here](http://wiki.ros.org/tf/Tutorials/Writing%20a%20tf%20listener%20%28Python%29) instead of [here](http://wiki.ros.org/tf2/Tutorials/Writing%20a%20tf2%20listener%20%28Python%29)

sr_multi_moveit/sr_multi_moveit_config/launch/trajectory_execution.launch.xml
```diff
@@ -13,7 +13,7 @@
   <!-- Allow more than the expected execution time before triggering a trajectory cancel (applied after scaling) -->
   <param name="trajectory_execution/allowed_goal_duration_margin" value="0.5"/> <!-- default 0.5 -->
   <!-- Allowed joint-value tolerance for validation that trajectory's first point matches current robot state -->
-  <param name="trajectory_execution/allowed_start_tolerance" value="0.03"/> <!-- default 0.01 -->
+  <param name="trajectory_execution/allowed_start_tolerance" value="6.0"/> <!-- default 0.01 -->
```
toliver, line 16: What function from the commander were you using to generate the plan/trajectory?
It seems that moveit checks the trajectory before sending it to the trajectory controller and it expects the first point to be in the current position.

toliver (changes requested):

beatrizleon:
I am closing this pr @ethanfowler considering toni's comments and it has conflicts now with the base.

PR #294, opened 2017-08-08
Added option to return plan, execute and check a specific plan

beatrizleon:
If none is passed it will work as before


AndriyPt:
@beatrizleon probably it will be better to create separate execute method with plan parameter. Or name it executeWithPlan or something like this. It looks a little bit too hacky to me :)

AndriyPt (approved):

PR #296, opened 2017-08-11
Fixing end effectors in the automatic generation process

beatrizleon:
As pointed out by a Guillaume's email, we need to fix the definition of the end effectors:

>  Dear Shadow team,
> 
> while fixing a reported issue on my dexterous manipulation tutorial 2 part 2, I noticed that the sr_multi_moveit package defines an end-effector that is invalid.
> 
> <end_effector group="right_arm" name="ra_moveit_ee" parent_link="ra_ee_link"/>
> 
> which tells that an end-effector named ra_moveit_ee, contains all the joints of the right_arm and is attached at the end of a kinematic chain at link ra_ee_link. The problem is that the right_arm group is not the end-effector, it is the kinematic chain one wants to plan the position of by using an end-effector.
> 
> <group name="right_arm">
>       <chain base_link="ra_base_link" tip_link="ra_ee_link"/>
>  </group>
> 
> In theory, the end-effector should not contain the joints of the arm, only one link is common between this kinematic group and this end-effector group. so for me it should be
> <end_effector group="right_hand" name="ra_moveit_ee" parent_link="ra_ee_link"/>
> the same for some others end-effectors.
> 
> I created a patch that I provide with my tutorial now
> https://github.com/guihomework/dexterous-manipulation-tutorial/blob/shadow_kinetic/srdfFix.patch
> 
> but it fixes only the end-effector of the right_arm, not the other end-effectors. I think the other end-effectors are more complex to generate (because one needs to create a group containing all other
> joints but the one of the parent group). For the arm_and_wrist group, this means its end-effector should contain the hand minus the wrist joints.

This pr contains modifications to solve the issue.
@guihomework could you confirm that the definition is now correct?

### DIFF
--- a/sr_moveit_hand_config/config/shadowhand.srdf.xacro
+++ b/sr_moveit_hand_config/config/shadowhand.srdf.xacro
@@ -17,13 +17,16 @@
   <xacro:macro name="fingerjoints" params="prefix joint_prefix lf">
     <xacro:if value="${lf}">
       <!-- Little finger special case -->
-     <joint name="${prefix}${joint_prefix}J5" /> 
-    </xacro:if> 
+     <joint name="${prefix}${joint_prefix}J5" />
+    </xacro:if>
     <joint name="${prefix}${joint_prefix}J4" />
     <joint name="${prefix}${joint_prefix}J3" />
     <joint name="${prefix}${joint_prefix}J2" />
     <joint name="${prefix}${joint_prefix}J1" />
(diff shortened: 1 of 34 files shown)


guihomework:
@beatrizleon  tested on rviz, everything is fine
Also tested in the dexterous tutorial which now works too.
thanks

beatrizleon:
Closes #300

AndriyPt (approved):

PR #302, opened 2017-09-04
Adding home starting positions for simulated ur10 and hand E.

ethanfowler:
Optional, and defaults to false. No change to the current launch interface, therefore shouldn't break anything.

### DIFF
--- a/sr_robot_launch/launch/sr_right_ur10arm_hand.launch
+++ b/sr_robot_launch/launch/sr_right_ur10arm_hand.launch
@@ -22,6 +22,8 @@
   <arg name="eth_port" default="$(optenv ETHERCAT_PORT eth0)"/>
   <!-- Set whether to load a scene to restrain the movement of the robot -->
   <arg name="scene" default="false"/>
+  <!-- Set to true if you want the robot to start in the "home" positions -->
+  <arg name="start_home" default="false"/>
 
   <!-- Set world value dependant on scene -->
   <arg name="world" default="$(find sr_description_common)/worlds/shadowhand.world" unless="$(arg scene)"/>
@@ -59,5 +61,6 @@
     <arg name="mapping_path" value="$(arg mapping_path)"/>
     <arg name="eth_port" value="$(arg eth_port)"/>
     <arg name="scene" value="$(arg scene)"/>
+    <arg name="start_home" default="$(arg start_home)"/>
   </include>
 </launch>
--- a/sr_robot_launch/launch/sr_right_ur10arm_hand.launch
+++ b/sr_robot_launch/launch/sr_right_ur10arm_hand.launch
@@ -24,6 +24,8 @@
   <arg name="scene" default="false"/>
   <!-- Set to true if you want the robot to start in the "home" positions -->
   <arg name="start_home" default="false"/>
+  <!-- Definition of "home" position, in joint-angle pairs -->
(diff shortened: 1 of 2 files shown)


sr_robot_launch/launch/sr_ur_arm_hand.launch
```diff
@@ -74,7 +76,9 @@
     <!-- HAND -->
     <param name="/hand/mapping/$(arg hand_serial)" value="$(arg hand_id)"/>
     <param name="/hand/joint_prefix/$(arg hand_serial)" value="$(arg hand_id)_"/>
-    <node name="spawn_model" pkg="gazebo_ros" type="spawn_model" args="-urdf -param robot_description -model $(arg robot_name)" respawn="false" output="screen"/>
+    <arg if="$(arg start_home)" name="spawn_model_extra_args" value="-unpause -J ra_shoulder_pan_joint 1.147 -J ra_elbow_joint 1.695 -J ra_wrist_1_joint -1.395 -J ra_wrist_2_joint -1.584 -J ra_shoulder_lift_joint -1.926 -J ra_wrist_3_joint 1.830" />
```
beatrizleon: I think we should have an arg "home_position" with these values as default but clearly defined so we can change those if we want from the ur10, etc launch file

ethanfowler: Agreed - I'll make the changes

ethanfowler: Changes made!

sr_robot_launch/launch/sr_ur_arm_hand.launch
```diff
@@ -21,6 +21,11 @@
   <arg name="world" default="$(find sr_description_common)/worlds/sr_lab.world" if="$(arg scene)"/>
   <!-- Set the scene file -->
   <arg name="scene_file" default="$(find sr_description_common)/scenes/lab.scene"/>
+  <!-- Set to true if you want the robot to start in the "home" positions -->
+  <arg name="start_home" default="false"/>
+  <!-- Definition of "home" position, in joint-angle pairs -->
+  <arg name="home_angles" default="-J ra_shoulder_pan_joint 1.147 -J ra_elbow_joint 1.695 -J ra_wrist_1_joint -1.395 -J ra_wrist_2_joint -1.584 -J ra_shoulder_lift_joint -1.926 -J ra_wrist_3_joint 1.830"/>
```
beatrizleon, line 27: probably the -J is not needed here and can be added later

sr_robot_launch/launch/sr_right_ur10arm_hand.launch
```diff
@@ -22,6 +22,8 @@
   <arg name="eth_port" default="$(optenv ETHERCAT_PORT eth0)"/>
   <!-- Set whether to load a scene to restrain the movement of the robot -->
   <arg name="scene" default="false"/>
+  <!-- Set to true if you want the robot to start in the "home" positions -->
+  <arg name="start_home" default="false"/>
```
beatrizleon, line 26: we should add the home position values here so we know where to change them

sr_robot_launch/launch/sr_ur_arm_hand.launch
ethanfowler: But there's a -J for every pair, which I can't think of any easy way around :/

sr_robot_launch/launch/sr_right_ur10arm_hand.launch
ethanfowler: Added!

beatrizleon (approved):

PR #309, opened 2017-09-20
F parameterise arm config

ethanfowler:
Parameterised the UR10 configuration file location, the UR10 program file location, and added launch files for two UR10s and one non-biotac Hand E (for testing).

All tested on HW and working.

### DIFF
--- a/sr_robot_launch/launch/sr_ur_arms_hand.launch
+++ b/sr_robot_launch/launch/sr_ur_arms_hand.launch
@@ -68,6 +68,9 @@
        More than one interface can be specified by concatenating them using underscore as a separator (e.g eth1_eth2_eth3) -->
   <arg name="eth_port" default="$(optenv ETHERCAT_PORT eth0)"/>
 
+  <!-- Robot description -->
+  <param name="robot_description" command="xacro --inorder $(arg robot_description) arm_1_z:=$(arg arm_1_z) arm_2_z:=$(arg arm_2_z) arm_separation:=$(arg arm_separation)"/>
+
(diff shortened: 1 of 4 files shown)


sr_robot_launch/launch/sr_ur_arms_hand.launch
```diff
@@ -117,15 +119,16 @@
...
+      <param name="ra_sr_ur_robot_hw/robot_program_path" type="str" value="$(arg arm_1_program_path)"/>
+      <param name="la_sr_ur_robot_hw/robot_program_path" type="str" value="$(arg arm_2_program_path)"/>
     </group>
 
     <rosparam unless="$(arg arm_ctrl)">
       robot_hardware:
         - unique_robot_hw
     </rosparam>
 
+    <param name="robot_description" command="xacro --inorder $(arg robot_description) arm_1_z:=$(arg arm_1_z) arm_2_z:=$(arg arm_2_z) arm_separation:=$(arg arm_separation)"/>
```
beatrizleon: if both sim and real need the robot description then it should be moved before the groups, instead of duplicate it

beatrizleon: But not sure why is not there in the single arm launch file for the real robot

beatrizleon (changes requested):

sr_robot_launch/launch/sr_ur_arms_hand.launch
ethanfowler: True! Fixed!

beatrizleon (approved):

PR #311, opened 2017-09-29
Improvements to the planner benchmarks 

beatrizleon:


.gitignore
```diff
@@ -39,3 +39,18 @@ CMakeLists.txt.user
...
+sr_moveit_hand_config/default_warehouse_mongo_db/
+sr_multi_moveit/sr_multi_moveit_config/default_warehouse_mongo_db/
+
+#Automatically generated files
+sr_moveit_hand_config/config/generated_shadowhand.srdf
+sr_moveit_hand_config/config/generated_shadowhand.urdf
+sr_multi_moveit/sr_multi_moveit_config/config/generated_robot.srdf
+sr_multi_moveit/sr_multi_moveit_config/config/controllers.yaml
+sr_multi_moveit/sr_multi_moveit_config/config/joint_limits.yaml
+sr_multi_moveit/sr_multi_moveit_config/config/kinematics.yaml
```
AndriyPt, line 54: @beatrizleon so as far as I understand we do not use these files at all?

sr_multi_moveit/sr_moveit_planner_benchmarking/launch/benchmarking.launch
```diff
@@ -1,6 +1,6 @@
 <launch>
   <arg name="planning" default="ompl"/>
-  <arg name="visualization" default="False"/>
+  <arg name="visualization" default="True"/>
```
AndriyPt: why have you turned on visualization? I guess for Jenkins it should be off by default?

ethanfowler (approved):

sr_multi_moveit/sr_multi_moveit_config/scripts/generate_load_moveit_config.py
```diff
@@ -47,7 +47,7 @@
 
     # detect the command to be executed
     if len(sys.argv) > 2:
-        save_file = False
+        save_file = True
```
AndriyPt: why have you changed this value? Does it influence Jenkins run?

AndriyPt (changes requested):

.gitignore
beatrizleon: we don't use them but these are the output of the automatically generated process. However they always appear when we do git status as untracked files and I thought it will be better to hide them in case someone want to commit them by mistake. Same for the mongo db folders

sr_multi_moveit/sr_moveit_planner_benchmarking/launch/benchmarking.launch
beatrizleon: True, I turned back off

sr_multi_moveit/sr_multi_moveit_config/scripts/generate_load_moveit_config.py
beatrizleon: We needed to have the files available for IKfast but as it is not part of this pr I turned back False

AndriyPt (approved):

PR #318, opened 2017-12-05
Add orientation constraints

giusebar:


sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -767,18 +768,22 @@ def change_teach_mode(mode, robot):
         except rospy.ServiceException:
             rospy.logerr("Failed to call service teach_mode")
 
-    def get_ik(self, target_pose, avoid_collisions=False, joint_states=None):
+    def get_ik(self, target_pose, ik_constraints=None, avoid_collisions=False, joint_states=None):
```
toliver: Let's make the new argument be the last argment, to avoid breaking the API.

toliver (changes requested):

toliver (approved):

PR #323, opened 2018-02-09
SRC-224 fixed trajectory planning

AndriyPt:


sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -392,7 +392,14 @@ def make_named_trajectory(self, trajectory):
 
         time_from_start = 0.0
 
-        for wp in trajectory:
+        current_waypoint = {
+            'joint_angles': [],
```
toliver, line 396: Quite a neat way of doing it!

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -392,7 +392,14 @@ def make_named_trajectory(self, trajectory):
...
+            'interpolate_time': 0.01,
```
toliver: Can we have a smaller interpolation time? (0.001s for example).
The idea is that this waypoint is as little disruptive as possible.

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -392,7 +392,14 @@ def make_named_trajectory(self, trajectory):
...
+            'pause_time': 0.01
```
toliver: Same here. We should make the pause_time=0.0 so that this waypoint doesn't introduce an unnecessary micro-stop of the robot.

toliver (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -392,7 +392,14 @@ def make_named_trajectory(self, trajectory):
 
         time_from_start = 0.0
 
-        for wp in trajectory:
+        current_waypoint = {
+            'joint_angles': [],
+            'interpolate_time': 0.001,
+            'pause_time': 0.0
+        }
+        waypoints = [current_waypoint] + trajectory
```
toliver: For this case, would it make sense to add a comment saying:
"We create a first trajectory point corresponding to the current position to avoid moveit complaining about it"

toliver (approved):

beatrizleon (approved):

PR #336, opened 2018-06-19
tested

dg-shadow:


sr_robot_commander/test/test_state_saver_no_hardware.test
```diff
@@ -0,0 +1,3 @@
+<launch>
```
AndriyPt, line 1: for unit test please use rosunit as here https://github.com/shadow-robot/build-servers-check/blob/kinetic-devel/sr_sample_python_library/test/test_sample_python_class.py

sr_robot_commander/src/sr_robot_commander/sr_robot_state_saver.py
```diff
@@ -20,15 +20,30 @@ def __init__(self, name, hand_or_arm="both"):
 
         self.__name = name
 
+        if name is None or name == '':
+            raise Exception("Cannot save with empty name.")
```
AndriyPt: Using Exception class for custom exceptions is not recommended since it is parent for so many exceptions and catch statement will catch everything. Use specific exception type in this case you could use ValueError https://docs.python.org/2/library/exceptions.html#exceptions.ValueError

AndriyPt (changes requested):

AndriyPt (approved):

PR #340, opened 2018-07-19
Fix name of gui arg in srhand.launch.

jackieokay:
Currently, setting the "gui" argument from the command line to false does not suppress the Rviz GUI launched by the moveit launch file. This pull request changes the argument to the correct name so that gui:=false works for an arbitrary launch file.


AndriyPt (approved):

AndriyPt:
@jacquelinekay thank you for fix :+1:

PR #341, opened 2018-08-15
Update hand commander to work with params for motor force controller

dg-shadow:
Need to wait for hand to test before merging


dg-shadow:
Tested

giusebar (approved):

giusebar (approved):

PR #429, opened 2019-11-06
F#src 3664 allow external control loop for hands

ghost:
## Proposed changes

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

￼


sr_robot_launch/launch/srhand_hardware_control_loop.launch
```diff
@@ -0,0 +1,79 @@
+<launch>
```
giusebar: I think this file is not needed, because we already have the https://github.com/shadow-robot/sr_interface/blob/kinetic-devel/sr_robot_launch/launch/sr_hardware_control_loop.launch, which has a variable called arm_ctrl. So, we could just call the same file but with arm_ctrl flag set to false

PR #446, opened 2020-01-08
F#src 3432 test integrate new driver arms hands

carebare47:
## Proposed changes

## Types of changes

## Checklist

- [ ] Read and follow the [contributing guidelines](https://github.com/shadow-robot/sr_documentation/blob/master/CONTRIBUTING.md).
￼


sr_robot_launch/launch/sr_hardware_control_loop.launch
```diff
@@ -124,9 +183,9 @@
       </group>
       <rosparam unless="$(arg hand_ctrl)" command="load" file="$(find sr_ur_launch)/config/ra_trajectory_controller.yaml"/>
       <node if="$(arg hand_ctrl)" name="arm_trajectory_controller_spawner" pkg="controller_manager" type="spawner" output="screen"
-            args="--wait-for=/$(arg arm_id)_arm_ready --wait-for=/calibrated $(arg arm_id)_trajectory_controller"/>
+            args="--wait-for=/$(arg arm_id)_sr_ur_robot_hw/robot_program_running $(arg arm_id)_trajectory_controller"/>
```
toliver, line 193: This will actually be an issue and it can't be merged.
We might need to modify the spawner code so that it accepts 2 topics to `wait-for`.
Remember that passing the argument twice, as we were doing, actually doesn't work (the second argument overwrites the first).
We could for example modify the spawner to allow for a string of comma-separated topics.
It would still have to work with a single topic, of course.

sr_robot_launch/launch/sr_hardware_control_loop.launch
```diff
@@ -40,6 +40,7 @@
   <!-- Set to true to spawn the position controllers for the arm-->
   <arg name="arm_position" default="$(eval not arm_trajectory)"/>
   <arg name="arm_speed_scale" default="0.5"/>
+  <arg name="kinematics_config" default="$(find sr_ur_calibration)/calibrations/bristol_ur10_calibration.yaml"/>
```
toliver: I think we don't want to introduce dependencies on sr_ur_calibration.
Shouldn't this be pointing to the default calibration?

toliver (changes requested):

sr_robot_launch/config/la_trajectory_controller.yaml
```diff
@@ -1,14 +1,15 @@
...
-  joints:
+hardware_interface:
+  joints: &robot_joints
     - la_shoulder_pan_joint
     - la_shoulder_lift_joint
     - la_elbow_joint
     - la_wrist_1_joint
     - la_wrist_2_joint
     - la_wrist_3_joint
-    - lh_WRJ1
```
biofotis, line 10: Why did you remove these joints?

sr_robot_launch/config/ra_trajectory_controller.yaml
```diff
@@ -1,14 +1,15 @@
...
-  joints:
+hardware_interface:
+  joints: &robot_joints
     - ra_shoulder_pan_joint
     - ra_shoulder_lift_joint
     - ra_elbow_joint
     - ra_wrist_1_joint
     - ra_wrist_2_joint
     - ra_wrist_3_joint
-    - rh_WRJ1
```
biofotis, line 10: here as well

sr_robot_launch/launch/sr_hardware_control_loop.launch
carebare47: Is this still the case since we'll be modifying joint_state_publisher.launch?

sr_robot_launch/launch/sr_hardware_control_loop.launch
carebare47: Yes, good spot https://github.com/shadow-robot/sr_interface/pull/446/commits/dfc48c2ad7b41e83a4c615b3f234232e4b8d25a3

sr_robot_launch/config/la_trajectory_controller.yaml
carebare47: Good question, should be fixed now

PR #465, opened 2020-02-14
F#src 3663 bimanual hands only

ghost:
## Proposed changes

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] Checked that all tests pass with my changes
- [x] Tested on real hardware (if appropriate)
￼

### DIFF
--- a/sr_robot_launch/launch/sr_bimanual.launch
+++ b/sr_robot_launch/launch/sr_bimanual.launch
@@ -21,7 +21,7 @@
   <!-- Set to true to spawn the position controllers for the hands-->
   <arg name="hand_ctrl" default="true"/>
   <!-- The control mode PWM (true) or torque (false) -->
-  <arg name="pwm_control" default="$(optenv PWM_CONTROL 1)"/>
+  <arg name="pwm_control" default="true"/>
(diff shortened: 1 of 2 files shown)


sr_robot_launch/launch/sr_bimanual.launch
```diff
@@ -21,7 +27,7 @@
   <!-- Set to true to spawn the position controllers for the hands-->
   <arg name="hand_ctrl" default="true"/>
   <!-- The control mode PWM (true) or torque (false) -->
-  <arg name="pwm_control" default="$(optenv PWM_CONTROL 1)"/>
+  <arg name="pwm_control" default="true"/>
```
beatrizleon, line 30: do we need to set this to true like this?

beatrizleon (approved):

sr_robot_launch/launch/sr_bimanual.launch
ghost: I did that because it was similar what we are doing in the unimanual launch file: https://github.com/shadow-robot/sr_interface/blob/melodic-devel/sr_robot_launch/launch/srhand.launch#L20

PR #511, opened 2020-06-10
using clean controller spawning

mikramarc:
## Proposed changes

## Types of changes

## Checklist

- [ ] Read and follow the [contributing guidelines](https://github.com/shadow-robot/sr_documentation/blob/master/CONTRIBUTING.md).
￼

### DIFF
--- a/sr_robot_launch/launch/srhand.launch
+++ b/sr_robot_launch/launch/srhand.launch
@@ -31,6 +31,8 @@
   <arg name="mapping_path" default="$(find sr_edc_launch)/mappings/default_mappings/$(arg hand_id)_E_v3.yaml"/>
   <!-- Set to true if you want to use grasp controller -->
   <arg name="grasp_controller" default="false"/>
+  <!-- Set to true if you want to use hybrid controller -->
+  <arg name="hybrid_controller" default="false"/>
   <!-- The robot state publisher publishing frequency  -->
   <arg name="robot_state_pub_frequency" default="250"/>
   <!-- The joint state publisher publishing frequency  -->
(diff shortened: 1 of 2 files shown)


sr_robot_launch/launch/srhand.launch
```diff
@@ -59,9 +61,10 @@
   <arg name="hand_ik_solver" default="sr_hand_kinematics"/>
 
   <!-- Default hand controller groups -->
+  <arg if="$(arg hybrid_controller)" name="hand_controller_group" default="hybrid"/>
   <arg if="$(arg grasp_controller)" name="hand_controller_group" default="grasp"/>
-  <arg if="$(eval hand_trajectory and not grasp_controller)" name="hand_controller_group" default="trajectory"/>
-  <arg if="$(eval not hand_trajectory and not grasp_controller)" name="hand_controller_group" default="position"/>
+  <arg if="$(eval hand_trajectory and not grasp_controller and not hybrid_controller)" name="hand_controller_group" default="trajectory"/>
+  <arg if="$(eval not hand_trajectory and not grasp_controller and not hybrid_controller)" name="hand_controller_group" default="position"/>
```
toliver, line 67: We need to have a discussion with the team about these arguments.
I think we should stop using them. They don't make sense anymore.
I think we should only keep the `hand_controller_group` argument. And the user will have to know the name of the group and pass it as an argument.

So, instead of `roslaunch srhand.launch hybrid_controller:=true` they will use `roslaunch srhand.launch hand_controller_group:=hybrid`

With the advantage that we get rid of all the `if` `unless` mess and we can add as many new types of controller groups as we want without having to modify the launchfile.

toliver (changes requested):

PR #537, opened 2020-08-06
Fixing controller spawning bug in which WRJ1+2 would not work when wr…

ethanfowler:
…ist was included in arm trajectory control


ethanfowler:
Tested in sim and HW in both haptx (include_wrist_in_arm_controller) and polhemus (not)

ghost (approved):

PR #586, opened 2021-01-06
fetch arm ips from param server

giusebar:
## Proposed changes

Fetch arm ips from param server
## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

Put an x in the boxes that apply. You can also fill these out after creating the PR. This is a reminder of what we should look for before merging this code.

- [x] Checked that all tests pass with my changes

### DIFF
--- a/sr_robot_launch/scripts/construct_robot_description
+++ b/sr_robot_launch/scripts/construct_robot_description
@@ -35,15 +35,19 @@ class SrConstructRobotDescription():
         CONST_ARM_TYPE = rospy.get_param("~arm_type")
         self._side = ''
         self._robot_description_params = {}
-        TIMEOUT_WAIT_FOR_PARAMS_IN_SECS = 60.0
 
-        self._right_arm_ip = '10.8.1.1'
-        self._left_arm_ip = '10.8.2.1'
+        TIMEOUT_IN_SECS = 60.0
+        right_arm_ip = '10.8.1.1'
+        left_arm_ip = '10.8.2.1'
 
-        self._wait_for_arm_ip_params(TIMEOUT_WAIT_FOR_PARAMS_IN_SECS)
+        if (self._arm_ip_params_loaded_before_timeout(TIMEOUT_IN_SECS)):
+            right_arm_ip, left_arm_ip = self._get_arm_ip_params()
+        else:
+            rospy.logwarn("No arm ip param was found on param server, using the following default arm ips:\n \
+                           right arm ip: {} - left arm ip: {}".format(right_arm_ip, left_arm_ip))
 
-        self._right_arm_params = {'prefix': 'ra', 'ip_address': self._right_arm_ip, 'arm_type': CONST_ARM_TYPE}
-        self._left_arm_params = {'prefix': 'la', 'ip_address': self._left_arm_ip, 'arm_type': CONST_ARM_TYPE}
+        self._right_arm_params = {'prefix': 'ra', 'ip_address': right_arm_ip, 'arm_type': CONST_ARM_TYPE}
+        self._left_arm_params = {'prefix': 'la', 'ip_address': left_arm_ip, 'arm_type': CONST_ARM_TYPE}
 
         if 'bimanual' in self._robot_description_file:
             self._side = 'both'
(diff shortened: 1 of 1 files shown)


sr_robot_launch/scripts/construct_robot_description
```diff
@@ -54,6 +61,18 @@ class SrConstructRobotDescription():
         rospy.set_param('robot_description', urdf)
         rospy.set_param('robot_description_ready', "true")
 
+    def _wait_for_arm_ip_params(self, timeout_in_secs):
+        start_time = rospy.get_time()
+        while not rospy.has_param("/la_sr_ur_robot_hw/robot_ip") and not rospy.has_param("/ra_sr_ur_robot_hw/robot_ip"):
+            if (rospy.get_time() - start_time > timeout_in_secs):
```
mykolasjuraitis, line 72: I wonder if it makes sense to sleep a little bit while waiting. Busy waiting will consume 100% CPU and further delay startup.

sr_robot_launch/scripts/construct_robot_description
```diff
@@ -35,8 +35,15 @@ class SrConstructRobotDescription():
...
-        self._CONST_RA_ARM = {'prefix': 'ra', 'ip_address': '10.8.1.1', 'arm_type': CONST_ARM_TYPE}
-        self._CONST_LA_ARM = {'prefix': 'la', 'ip_address': '10.8.2.1', 'arm_type': CONST_ARM_TYPE}
+        TIMEOUT_WAIT_FOR_PARAMS_IN_SECS = 60.0
+
+        self._right_arm_ip = '10.8.1.1'
+        self._left_arm_ip = '10.8.2.1'
+
+        self._wait_for_arm_ip_params(TIMEOUT_WAIT_FOR_PARAMS_IN_SECS)
+
+        self._right_arm_params = {'prefix': 'ra', 'ip_address': self._right_arm_ip, 'arm_type': CONST_ARM_TYPE}
```
mykolasjuraitis: You are changing _CONST_RA_ARM -> _right_arm_params here but the rest of the code still uses old name _CONST_RA_ARM.

sr_robot_launch/scripts/construct_robot_description
```diff
@@ -35,8 +35,15 @@ class SrConstructRobotDescription():
...
         self._side = ''
         self._robot_description_params = {}
-        self._CONST_RA_ARM = {'prefix': 'ra', 'ip_address': '10.8.1.1', 'arm_type': CONST_ARM_TYPE}
-        self._CONST_LA_ARM = {'prefix': 'la', 'ip_address': '10.8.2.1', 'arm_type': CONST_ARM_TYPE}
+        TIMEOUT_WAIT_FOR_PARAMS_IN_SECS = 60.0
+
+        self._right_arm_ip = '10.8.1.1'
+        self._left_arm_ip = '10.8.2.1'
+
+        self._wait_for_arm_ip_params(TIMEOUT_WAIT_FOR_PARAMS_IN_SECS)
```
mykolasjuraitis: Name `_wait_for_arm_ip_params` implies waiting only but it also has side effect of changing _right_arm_ip and _left_arm_ip which are used for this function call only. Wouldn't it be better to do either:
1. Make function _wait_for_arm_ip_params return only boolean if it timed out or not and then get parameters directly here in constructor or use default values thus making _right_arm_ip and _left_arm_ip local variables.
2. Make function _wait_for_arm_ip_params take 2 additional parameters of default IP values and return 2 results which will be either real parameters or defaults depending on timeout.

mykolasjuraitis:
I think this line in the PR description:

> Please explain the changes you made. Add pictures or videos to explain them better if appropriate.

should be replaced with an actual description of changes. For example:

> Fetch arm ips from param server.

sr_robot_launch/scripts/construct_robot_description
giusebar: addressed in https://github.com/shadow-robot/sr_interface/pull/586/commits/df44fa98bba485f979ea12e3ef69c766f53209fa

sr_robot_launch/scripts/construct_robot_description
giusebar: addressed in https://github.com/shadow-robot/sr_interface/pull/586/commits/df44fa98bba485f979ea12e3ef69c766f53209fa

sr_robot_launch/scripts/construct_robot_description
giusebar: addressed in https://github.com/shadow-robot/sr_interface/pull/586/commits/df44fa98bba485f979ea12e3ef69c766f53209fa

mykolasjuraitis (approved):

ghost (approved):

PR #597, opened 2021-01-26
F#src 5451 arms braking

giusebar:
## Proposed changes

Add functionality to brake UR10s when needed. 

## Checklist

- [x] Manually tested that added code works as intended (if any functional/runnable code is added).
- [x] Added automated tests (if a new class is added (Python or C++), interface of that class must be unit tested).
- [x] Tested on real hardware (if the changed or added code is used with the real hardware).
- [x] Added documentation (For any new feature, explain what it does and how to use it. Write the documentation in a relevant space, e.g. Github, Confluence, etc.)


sr_robot_launch/scripts/sr_ur_arm_unlock
```diff
@@ -31,11 +32,34 @@ class SrUrUnlock():
...
+            list_controllers = rospy.ServiceProxy('controller_manager/list_controllers', ListControllers)
+            loaded_controller = list_controllers()
+        except rospy.ServiceException, e:
+            rospy.logerr("Could not get controllers list: %s", e)
+
+        for controller in loaded_controller.controller:
+            if arm + "_trajectory_controller" in controller.name:
+                if controller.state == "running":
+                    controller_is_running = True
+                    return controller_is_running
```
mykolasjuraitis: Given method name `is_arm_controller_running` it kind of feels it should always return boolean value but now it will return either `True` or `None`. Also if `ServiceException` is thrown wouldn't `loaded_controller` be undefined so should not be used in the loop? Maybe better to have something like:
```
        try:
            list_controllers = rospy.ServiceProxy('controller_manager/list_controllers', ListControllers)
            loaded_controller = list_controllers()
            for controller in loaded_controller.controller:
                if arm + "_trajectory_controller" in controller.name:
                    if controller.state == "running":
                        return True
        except rospy.ServiceException, e:
            rospy.logerr("Could not get controllers list: %s", e)
        return False
```

mykolasjuraitis (approved):

biofotis (approved):

sr_robot_launch/scripts/sr_ur_arm_unlock
giusebar: wow, really didn't put enough attention here, thanks for review. addressed in https://github.com/shadow-robot/sr_interface/pull/597/commits/cfb498354aad156c2652b626e9d80cc49ec0fb91

PR #611, opened 2021-03-02
SRC-5987 Publish underactuation error

mykolasjuraitis:
## Proposed changes

Calculate and Publish underactuation error.

The following new topics are added:
* /underactuation_error/rh_fftip
* /underactuation_error/rh_lftip
* /underactuation_error/rh_mftip
* /underactuation_error/rh_rftip

Example of new message:
```
header: 
  seq: 2359
  stamp: 
    secs: 32
    nsecs:  39000000
  frame_id: ''
error: 0.0123959742486
```

New node is included in launch files `srhand.launch`, `sr_ur_arms_hand.launch` and `sr_ur_arms_hands.launch` or can be started manualy by running:
```
rosrun sr_error_reporter error_reporter
```

## Checklist

- [x] Manually tested that added code works as intended (if any functional/runnable code is added).
- [x] Added automated tests (if a new class is added (Python or C++), interface of that class must be unit tested).
- [x] Tested on real hardware (if the changed or added code is used with the real hardware).
- [x] Added documentation (For any new feature, explain what it does and how to use it. Write the documentation in a relevant space, e.g. Github, Confluence, etc.)

### DIFF
--- a/sr_error_reporter/README.md
+++ b/sr_error_reporter/README.md
@@ -17,7 +17,11 @@ only J1 and J2 joints for underactuated fingers are set to either actual or
 desired values. Then fingertip position is calculated by running forward
 kinematics. Cartesian distance between actual and desired fingertips is
 calculated and published to a topic per finger:
-* /underactuation_error/rh_fftip
-* /underactuation_error/rh_lftip
-* /underactuation_error/rh_mftip
-* /underactuation_error/rh_rftip
+* /sh_lh_ffj0_position_controller/underactuation_cartesian_error
+* /sh_lh_mfj0_position_controller/underactuation_cartesian_error
+* /sh_lh_rfj0_position_controller/underactuation_cartesian_error
+* /sh_lh_lfj0_position_controller/underactuation_cartesian_error
+* /sh_rh_ffj0_position_controller/underactuation_cartesian_error
+* /sh_rh_mfj0_position_controller/underactuation_cartesian_error
+* /sh_rh_rfj0_position_controller/underactuation_cartesian_error
+* /sh_rh_lfj0_position_controller/underactuation_cartesian_error
--- a/sr_error_reporter/README.md
+++ b/sr_error_reporter/README.md
@@ -9,14 +9,14 @@ rosrun sr_error_reporter error_reporter
 
 ## Underactuation error
 
-Underactuation error is calculated by listening to topic `/joint_states` for
-actual values from sensors and by listening to topics `/lh_trajectory_controller/command`
-and `/rh_trajectory_controller/command` for desired values.
-When one of these topics changes, kinematic model is reset to zeros and only J1
-and J2 joints for underactuated fingers are set to either actual or desired
-values. Then fingertip position is calculated by running forward kinemates.
-Cartesian distance between actual and desired fingertips are calculated and
-published to a topic per finger:
+Underactuation error is calculated by listening to topics
+`/lh_trajectory_controller/state` and `/rh_trajectory_controller/state` and
+comparing `actual.positions` field to `desired.positions` field.
+When one of these topics changes, the kinematic model is reset to zeros and
+only J1 and J2 joints for underactuated fingers are set to either actual or
+desired values. Then fingertip position is calculated by running forward
+kinematics. Cartesian distance between actual and desired fingertips is
+calculated and published to a topic per finger:
 * /underactuation_error/rh_fftip
 * /underactuation_error/rh_lftip
 * /underactuation_error/rh_mftip
--- a/sr_error_reporter/README.md
+++ b/sr_error_reporter/README.md
@@ -1,0 +1,23 @@
+# sr_error_reporter
+
+A package for calculating and publishing errors.
+
+This node can be started by:
+```
+rosrun sr_error_reporter error_reporter
+```
+
+## Underactuation error
+
+Underactuation error is calculated by listening to topic `/joint_states` for
+actual values from sensors and by listening to topics `/lh_trajectory_controller/command`
+and `/lh_trajectory_controller/command` for desired values.
+When one of these topics changes, kinematic model is reset to zeros and only J1
+and J2 joints for underactuated fingers are set to either actual or desired
+values. Then fingertip position is calculated by running forward kinemates.
+Cartesian distance between actual and desired fingertips are calculated and
+published to a topic per finger:
+* /underactuation_error/rh_fftip
+* /underactuation_error/rh_lftip
+* /underactuation_error/rh_mftip
+* /underactuation_error/rh_rftip
--- a/sr_error_reporter/README.md
+++ b/sr_error_reporter/README.md
@@ -11,7 +11,7 @@ rosrun sr_error_reporter error_reporter
 
 Underactuation error is calculated by listening to topic `/joint_states` for
 actual values from sensors and by listening to topics `/lh_trajectory_controller/command`
-and `/lh_trajectory_controller/command` for desired values.
+and `/rh_trajectory_controller/command` for desired values.
 When one of these topics changes, kinematic model is reset to zeros and only J1
 and J2 joints for underactuated fingers are set to either actual or desired
 values. Then fingertip position is calculated by running forward kinemates.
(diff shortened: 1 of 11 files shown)


sr_error_reporter/README.md
```diff
@@ -0,0 +1,23 @@
...
+This node can be started by:
+```
+rosrun sr_error_reporter error_reporter
+```
+
+## Underactuation error
+
+Underactuation error is calculated by listening to topic `/joint_states` for
+actual values from sensors and by listening to topics `/lh_trajectory_controller/command`
+and `/lh_trajectory_controller/command` for desired values.
```
maxzieba: /rh_trajectory_controller/command

mykolasjuraitis: Well spotted! Fixed in 80091116827c6404158c8d71d1bc4903c4906bf5.

sr_error_reporter/src/UnderactuationErrorReporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include "eigen_conversions/eigen_msg.h"
+#include "std_msgs/String.h"
+#include "sr_error_reporter/UnderactuationError.h"
+#include "sr_error_reporter/UnderactuationErrorReporter.hpp"
+
+UnderactuationErrorReporter::UnderactuationErrorReporter(ros::NodeHandle& node_handle)
+  : node_handle_(node_handle)
+{
+  joints_subsriber_ = node_handle.subscribe("/joint_states", 1,
```
giusebar: typo in subscriber naming

sr_error_reporter/src/UnderactuationErrorReporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+  {
+    for (auto& finger : include_fingers_)
+    {
+      auto j1 = joint_positions.find(side + finger.first + "distal");
+      auto j2 = joint_positions.find(side + finger.first + "middle");
+      std::vector<double> positions = {0, 0, j2->second, j1->second};
+      robot_state_->setJointGroupPositions(side + finger.second, positions);
+    }
+  }
+  for (auto& side : sides_)
```
giusebar: why do we do this in a separate loop, can't we do it in the loop above?

sr_error_reporter/src/UnderactuationErrorReporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+}
+
+void UnderactuationErrorReporter::update_joint_position(
+  std::map<std::string, double>& joint_positions,
+  std::string name,
+  double position)
+{
+  if (name.size() < 7)
```
giusebar, line 125: what is 7? maybe add a const?

sr_error_reporter/src/UnderactuationErrorReporter.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+ros::Publisher UnderactuationErrorReporter::get_or_create_publisher(std::string link_name)
+{
+  auto iterator = error_publishers_.find(link_name);
+  if (iterator != error_publishers_.end())
+  {
+    return iterator->second;
+  }
+  ros::Publisher publister = node_handle_.advertise<sr_error_reporter::UnderactuationError>(
```
giusebar: publisher typo

giusebar (changes requested):

sr_error_reporter/include/sr_error_reporter/UnderactuationErrorReporter.hpp
```diff
@@ -0,0 +1,116 @@
...
+   * MoveIt link name to radians. From /[lh|rh]_trajectory_controller/command topic.
+   */
+  std::map<std::string, double> desired_joint_angles_;
+
+  std::map<std::string, geometry_msgs::Transform> actual_tip_transforms_;
+  std::map<std::string, geometry_msgs::Transform> desired_tip_transforms_;
+
+  std::set<std::string> sides_;
+
+  std::map<std::string, std::string> include_fingers_ =
```
giusebar, line 42: we need to be careful when setting this in launch file. Lite hand does not have ring finger and little finger, not sure if we are interested in checking the error there, But we could protect against that when we look for the joint in the loop in line 64.

sr_error_reporter/src/UnderactuationErrorReporter.cpp
mykolasjuraitis: Renamed in bf3029a16310672b6159f2cfc6a2e764489858e1.

sr_error_reporter/src/UnderactuationErrorReporter.cpp
mykolasjuraitis: Renamed in bf3029a16310672b6159f2cfc6a2e764489858e1.

sr_error_reporter/src/UnderactuationErrorReporter.cpp
mykolasjuraitis: Added a comment in bf3029a16310672b6159f2cfc6a2e764489858e1 with a example of possible name. Hopefully that makes it clear which length is expected.

sr_error_reporter/src/UnderactuationErrorReporter.cpp
mykolasjuraitis: Your are right. Initially I was not sure in which order I will update kinematics model, i.e. by fingers or by joints, but, since it is by fingers now, I merged loops in bf3029a16310672b6159f2cfc6a2e764489858e1.

sr_error_reporter/include/sr_error_reporter/UnderactuationErrorReporter.hpp
mykolasjuraitis: It might be a bit confusing but this map purpose it to exclude thumb and wrist joints. If ring finger and little finger are not present in /joint_states or /[lh|rh]_trajectory_controller/command they will be ignored too. So errors for ring and little fingers will be published only if lite hand publishes commands and joint states for them. Does it?

sr_error_reporter/src/UnderactuationErrorReporter.cpp
```diff
@@ -0,0 +1,168 @@
...
+  while (ros::ok())
+  {
+    std::string robot_description;
+    if (node_handle.searchParam("robot_description_semantic", robot_description))
+    {
+      robot_model_loader_.reset(new robot_model_loader::RobotModelLoader("robot_description"));
+      robot_state_.reset(new robot_state::RobotState(robot_model_loader_->getModel()));
+      break;
+    }
+    ROS_INFO("Waiting for robot description");
```
giusebar: ah we should wait for the parameter to be available actually, for reference look here, we have implemented a wait_for_param utility for cases like this one ->https://github.com/shadow-robot/jog_arm_internal/blob/f8dda4acd3ff02f006d2948fafb136a693b3bb4e/jog_arm/src/jog_arm/jog_arm_server.cpp#L98-L102

giusebar (changes requested):

sr_error_reporter/src/UnderactuationErrorReporter.cpp
mykolasjuraitis: Changed to use `wait_for_param` utility in bb8e6bc7b845fad190f9e568422db93d544fc863.

giusebar (approved):

giusebar (approved):

ethanfowler:
Should the published topic names be more specifically Shadow? e.g. "sh_underactuation_error"? Or nested in an existing topic tree, e.g. "/sh_rh_rfj0_position_controller/underactuation_error", or "rh_trajectory_controller/underactuation_error"? Thoughts @toliver ?

toliver:
> Should the published topic names be more specifically Shadow? e.g. "sh_underactuation_error"? Or nested in an existing topic tree, e.g. "/sh_rh_rfj0_position_controller/underactuation_error", or "rh_trajectory_controller/underactuation_error"? Thoughts @toliver ?

`/sh_rh_rfj0_position_controller/underactuation_error` could be a good option, but I don't have a strong preference for it. We could also put them all under the same namespace.
What I think would be helpful is to specify that we're talking about cartesian error. This is obvious to us now, but maybe it's better if it's part of the name.

`/sh_rh_rfj0_position_controller/underactuation_cartesian_error`

biofotis (approved):

PR #649, opened 2021-06-22
Tests for SrRobotCommander

maxzieba:
## Proposed changes

Adding tests for robot_commander and minor fixes.

## Checklist

- [X] Manually tested that added code works as intended (if any functional/runnable code is added).
- [X] Added automated tests (if a new class is added (Python or C++), interface of that class must be unit tested).
- [ ] ~~Tested on real hardware (if the changed or added code is used with the real hardware).~~
- [ ] ~~Added documentation (For any new feature, explain what it does and how to use it. Write the documentation in a relevant space, e.g. Github, Confluence, etc.)~~


sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -246,7 +246,7 @@ def plan_to_joint_value_target(self, joint_states, angle_degrees=False, custom_s
         else:
             self._move_group_commander.set_start_state(custom_start_state)
         self._move_group_commander.set_joint_value_target(joint_states_cpy)
-        self.__plan = self._move_group_commander.plan()
+        self.__plan = self._move_group_commander.plan()[1]
```
giusebar: just to make it clearer for reader maybe we should set 1 to const variable. e.g. TUPLE_TRAJECTRORY_INDEX = 1

giusebar: and then use it in all the functions .and maybe above the const explain the change

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,466 @@
...
+        cls.robot_commander = SrRobotCommander("right_arm")
+
+    def test_mvgr_get_and_set_planner_id(self):
+        planner_list = ["BKPIECEkConfigDefault", "ESTkConfigDefault", "KPIECEkConfigDefault",
+                        "LBKPIECEkConfigDefault", "PRMkConfigDefault", "RRTkConfigDefault",
+                        "SBLkConfigDefault", "PRMstarkConfigDefault", "RRTstarkConfigDefault"]
+
+        for id in planner_list:
+            self.robot_commander.set_planner_id(id)
+            self.assertTrue(id == self.robot_commander._move_group_commander.get_planner_id())
```
giusebar: will this just fail at the first failure? or execute all the calls and report each fail? this might be bad in case some id gets deprecated and some do not. I think we should use https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,466 @@
...
+
+    def test_mvgr_set_num_planning_attempts(self):
+        raised = False
+        try:
+            self.robot_commander.set_num_planning_attempts(1)
+        except (NameError, TypeError):
+            raised = True
+        self.assertFalse(raised)
+
+    # def get_end_effector_pose_from_named_state(self, name):
```
giusebar: remove

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,466 @@
...
+                 'ra_elbow_joint': 0.6876824920327893,
+                 'ra_wrist_1_joint': -0.7695210732233582,
+                 'ra_wrist_2_joint': 0.2298871642157314,
+                 'ra_shoulder_lift_joint': -0.9569080092786892,
+                 'ra_wrist_3_joint': -0.25991215955733704}
+        rs = RobotState()
+        for key, value in state.items():
+            rs.joint_state.name.append(key)
+            rs.joint_state.position.append(value)
+        pose = self.robot_commander.get_end_effector_pose_from_state(rs)
```
giusebar: add assertEqual pose expected ? or at least that pose is not empty and it's a valid number

giusebar: or we should check what happens if state does not return a valid solution? is that possible?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,466 @@
...
+        try:
+            self.robot_commander.get_planning_frame()
+        except (NameError, TypeError):
+            raised = True
+        self.assertFalse(raised)
+
+    def test_mvgr_set_and_get_pose_reference_frame(self):
+        reference_frame = "world"
+        self.robot_commander._move_group_commander.set_pose_reference_frame(reference_frame)
+        self.assertTrue(self.robot_commander._move_group_commander.get_pose_reference_frame() == reference_frame)
```
giusebar: use assertEqual

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,466 @@
...
+
+    def test_get_group_name(self):
+        self.assertTrue(self.robot_commander._name == self.robot_commander.get_group_name())
```
giusebar: assertequal

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+            self.robot_commander.set_max_velocity_scaling_factor(0.2)
+        except Exception:
+            raised = True
+        self.assertFalse(raised)
+
+    def test_set_max_velocity_scaling_factor__range_not_ok(self):
+        raised = False
+        try:
+            self.robot_commander.set_max_velocity_scaling_factor(3)
+        except Exception:
```
giusebar: here you can just have assertRaises if you know the type of exception.

e.g. self.assertRaises(ExpectedException, self.robot_commander.set_max_velocity_scaling_factor, 3)

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+        self.robot_commander.refresh_named_targets()
+        condition_1 = type(self.robot_commander._srdf_names) == list
+        condition_2 = type(self.robot_commander._warehouse_names) == list
+        self.assertTrue(condition_1 and condition_2)
+
+    def test_set_max_velocity_scaling_factor__range_ok(self):
+        raised = False
+        try:
+            self.robot_commander.set_max_velocity_scaling_factor(0.2)
+        except Exception:
```
giusebar: i think you don't need to wrap up this in a try catch, the unit test will fail if the exception is thrown.

it can just be self.robot_commander.set_max_velocity_scaling_factor(0.2). and should pass. maybe add a comment like /*expected no exception to be raised*/. Up to you tho, not entirely sure about this one :)

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+        try:
+            self.robot_commander.set_max_acceleration_scaling_factor(0.2)
+        except Exception:
+            raised = True
+        self.assertFalse(raised)
+
+    def test_set_max_acceleration_scaling_factor__range_not_ok(self):
+        raised = False
+        try:
+            self.robot_commander.set_max_acceleration_scaling_factor(3)
```
giusebar: same as above

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+        self.robot_commander.allow_looking(True)
+
+    def test_allow_replanning(self):
+        self.robot_commander.allow_replanning(True)
+
+    def test_plan_to_joint_value_target(self):
+        arm_home_joints_goal = {'ra_shoulder_pan_joint': 0.00, 'ra_elbow_joint': 2.00,
+                                'ra_shoulder_lift_joint': -1.25, 'ra_wrist_1_joint': -0.733,
+                                'ra_wrist_2_joint': 1.5708, 'ra_wrist_3_joint': 0.00}
+        self.robot_commander.plan_to_joint_value_target(arm_home_joints_goal,
```
giusebar: this function returns a plan, we should use the return of the function to check. I think it would be good to check if the last point of the plan corresponds with the goal you set, more than just seeing if it's a robot_trajectory message.

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+        self.robot_commander.execute()
+        executed_joints = self.robot_commander.get_current_state()
+        diffs = [arm_home_joints_goal[key] - executed_joints[key] for key in arm_home_joints_goal
+                 if key in executed_joints]
+        condition_1 = all(diff < 0.05 for diff in diffs)
+        condition_2 = self.robot_commander._SrRobotCommander__plan is None
+        self.assertTrue(condition_1 and condition_2)
+
+    def test_execute_plan(self):
+        arm_home_joints_goal = {'ra_shoulder_pan_joint': 0.00, 'ra_elbow_joint': 2.00,
```
giusebar: if we use always the same joint angles maybe we should reset the arm position for every test, maybe you can have a common function that plans and executes to home position. OR you can use different positions.

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+                                'ra_shoulder_lift_joint': -1.25, 'ra_wrist_1_joint': -0.733,
+                                'ra_wrist_2_joint': 1.5708, 'ra_wrist_3_joint': 0.00}
+        self.robot_commander.plan_to_joint_value_target(arm_home_joints_goal,
+                                                        angle_degrees=False, custom_start_state=None)
+        self.robot_commander.execute()
+        executed_joints = self.robot_commander.get_current_state()
+        diffs = [arm_home_joints_goal[key] - executed_joints[key] for key in arm_home_joints_goal
+                 if key in executed_joints]
+        condition_1 = all(diff < 0.05 for diff in diffs)
+        condition_2 = self.robot_commander._SrRobotCommander__plan is None
```
giusebar: why is this condition needed?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+    def test_evaluate_given_plan__high_quality(self):
+        current_joints = self.robot_commander.get_joints_position()
+        planned_joints = copy.deepcopy(current_joints)
+        planned_joints['ra_shoulder_pan_joint'] = planned_joints['ra_shoulder_pan_joint'] + 0.001
+        plan = self.robot_commander.plan_to_joint_value_target(planned_joints, angle_degrees=False,
+                                                               custom_start_state=None)
+        evaluation = self.robot_commander.evaluate_given_plan(plan)
+        self.assertTrue(evaluation < 1)
+
+    # def test_evaluate_plan(self):
```
giusebar: remove

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+        self.assertTrue(evaluation > 10)
+
+    def test_evaluate_given_plan__high_quality(self):
+        current_joints = self.robot_commander.get_joints_position()
+        planned_joints = copy.deepcopy(current_joints)
+        planned_joints['ra_shoulder_pan_joint'] = planned_joints['ra_shoulder_pan_joint'] + 0.001
+        plan = self.robot_commander.plan_to_joint_value_target(planned_joints, angle_degrees=False,
+                                                               custom_start_state=None)
+        evaluation = self.robot_commander.evaluate_given_plan(plan)
+        self.assertTrue(evaluation < 1)
```
giusebar: use assertLess(a, b) -> a < b

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+
+    def test_evaluate_given_plan__low_quality(self):
+        current_joints = self.robot_commander.get_joints_position()
+        planned_joints = copy.deepcopy(current_joints)
+        planned_joints['ra_shoulder_pan_joint'] = planned_joints['ra_shoulder_pan_joint'] + 3.14
+        planned_joints['ra_shoulder_lift_joint'] = planned_joints['ra_shoulder_lift_joint'] + 1.36
+        plan = self.robot_commander.plan_to_joint_value_target(planned_joints, angle_degrees=False,
+                                                               custom_start_state=None)
+        evaluation = self.robot_commander.evaluate_given_plan(plan)
+        self.assertTrue(evaluation > 10)
```
giusebar: use assertGreater(a, b) -> a>b

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,465 @@
...
+                                                               angle_degrees=False, custom_start_state=rs)
+        self.robot_commander.execute_plan(plan)
+        rospy.sleep(0.5)
+        self.robot_commander.send_stop_trajectory_unsafe()
+        current_joints = self.robot_commander.get_current_state()
+
+        diffs = [end_joints[key] - current_joints[key] for key in end_joints
+                 if key in current_joints]
+
+        rospy.logerr(diffs)
```
giusebar: remove logerr

giusebar (changes requested):

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: condition_1 is just to accept slight offset in final position after plan execution

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Removed other planner id's, the setter&getter are just setting the id, but internally within moveit it's not verified. This test is rather to check if API will change

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: left bare method call

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -44,6 +44,9 @@
 import copy
 import numpy
 
+# Since Moveit update to noetic, the plan() method returns a tuple where trajectory is indexed at 1. 
+# More info here: https://github.com/ros-planning/moveit/blob/master/MIGRATION.md
```
giusebar, line 48: nice :+1:

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,458 @@
...
+
+class TestSrRobotCommander(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.robot_commander = SrRobotCommander("right_arm")
+
+    def test_mvgr_get_and_set_planner_id(self):
+        planner_id = "BKPIECEkConfigDefault"
+        self.robot_commander.set_planner_id(planner_id)
+        #st = planner_id + " " + str(self.robot_commander._move_group_commander.get_planner_id())
```
giusebar: remove comment

sr_robot_commander/test/test_robot_commander.py
giusebar: I meant why plan has to be None

giusebar (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
mikramarc: nice indeed!

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -44,6 +45,10 @@
 import copy
 import numpy
 
+# Since Moveit update to noetic, the plan() method returns a tuple where trajectory is indexed at 1.
+# More info here: https://github.com/ros-planning/moveit/blob/master/MIGRATION.md
+TUPLE_TRAJECTORY_INDEX = 1
```
mikramarc: please add CONST_ prefix as is our convention

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -661,29 +665,41 @@ def run_named_trajectory(self, trajectory):
...
         @param wait - should method wait for movement end or not.
         """
+        pose = self._move_group_commander.get_current_pose()
+        pose.pose.position.x = xyz[0]
+        pose.pose.position.y = xyz[1]
+        pose.pose.position.z = xyz[2]
+
         self._move_group_commander.set_start_state_to_current_state()
-        self._move_group_commander.set_position_target(xyz, end_effector_link)
+        self._move_group_commander.set_pose_target(pose, end_effector_link)
```
mikramarc, line 679: why not just use position target?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -661,29 +665,41 @@ def run_named_trajectory(self, trajectory):
...
+        pose.pose.position.y = xyz[1]
+        pose.pose.position.z = xyz[2]
+
         if custom_start_state is None:
             self._move_group_commander.set_start_state_to_current_state()
         else:
             self._move_group_commander.set_start_state(custom_start_state)
-        self._move_group_commander.set_position_target(xyz, end_effector_link)
-        self.__plan = self._move_group_commander.plan()[1]
+        self._move_group_commander.set_pose_target(pose, end_effector_link)
```
mikramarc, line 699: again, why not position target?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+from std_msgs.msg import Header
+from sensor_msgs.msg import JointState
+from moveit_commander.exception import MoveItCommanderException
+from moveit_commander import conversions
+from actionlib_msgs.msg import GoalStatusArray
+import tf2_ros
+import time
+
+PKG = "sr_robot_commander"
+RA_HOME_ANGLES = {'ra_shoulder_pan_joint': 0.00, 'ra_elbow_joint': 2.00,
```
mikramarc: CONST

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+                  'ra_shoulder_lift_joint': -1.57, 'ra_wrist_1_joint': -0.73,
+                  'ra_wrist_2_joint': 1.57, 'ra_wrist_3_joint': 0.00}
+
+RA_EXAMPLE_TARGET = {'ra_shoulder_pan_joint': 0.2, 'ra_elbow_joint': 1.80,
```
mikramarc: CONST

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+        plan = self.robot_commander.plan_to_joint_value_target(RA_HOME_ANGLES, angle_degrees=False,
+                                                               custom_start_state=None)
+        self.robot_commander.execute_plan(plan)
+
+    def compare_poses(self, pose1, pose2, tolerance=0.01):
+        p1_list = [pose1.position.x, pose1.position.y, pose1.position.z,
+                   pose1.orientation.x, pose1.orientation.y, pose1.orientation.z, pose1.orientation.w]
+        p2_list = [pose2.position.x, pose2.position.y, pose2.position.z,
+                   pose2.orientation.x, pose2.orientation.y, pose2.orientation.z, pose2.orientation.w]
+        digit = 2
```
mikramarc: round is a well known function, I'd say no need for additional variable

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+        p1_list = [round(i, digit) for i in p1_list]
+        p2_list = [round(i, digit) for i in p2_list]
+
+        for idx, vals in enumerate(zip(p1_list, p2_list)):
+            if vals[0] == -0.0:
```
mikramarc: why is this casting to integer necessary?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+                p2_list[idx] = 0
+            if abs(p1_list[idx]) - abs(p2_list[idx]) > tolerance:
+                return False
+        return True
+
+    def compare_joint_states(self, js1, js2, tolerance=0.01):
+        common_keys = set(js1.keys()).intersection(set(js2.keys()))
+        if len(common_keys) == 0:
+            return False
+        else:
```
mikramarc: no need for this else

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+            digit = 2
```
mikramarc: remove

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+        for idx, vals in enumerate(zip(p1_list, p2_list)):
+            if vals[0] == -0.0:
+                p1_list[idx] = 0
+            if vals[1] == -0.0:
+                p2_list[idx] = 0
+            if abs(p1_list[idx]) - abs(p2_list[idx]) > tolerance:
+                return False
+        return True
+
+    def compare_joint_states(self, js1, js2, tolerance=0.01):
```
mikramarc: please just name joint_state_1 etc, current makes it less readable when you go down

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+        common_keys = set(js1.keys()).intersection(set(js2.keys()))
```
mikramarc: hm, so are we considering joint states the same if they have all keys different apart from one? is that desired?

mikramarc: also, instead of keys, can we say common_joint_names or something like that?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+            digit = 2
+            for key in common_keys:
+                js1[key] = round(js1[key], digit)
+                js2[key] = round(js2[key], digit)
+                if abs(js1[key]) - abs(js2[key]) > tolerance:
+                    return False
+        return True
+
+    def test_get_and_set_planner_id(self):
+        planner_id = "RRTstarkConfigDefault"
```
mikramarc: const

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+        self.robot_commander.set_planner_id(planner_id)
+        self.assertEqual(planner_id, self.robot_commander._move_group_commander.get_planner_id())
+
+    def test_get_and_set_planning_time(self):
+        time_test_value = 3
```
mikramarc: const

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+        self.robot_commander.set_planning_time(time_test_value)
+        self.assertEqual(time_test_value, self.robot_commander._move_group_commander.get_planning_time())
+
+    def test_set_num_planning_attempts(self):
+        raised = False
+        try:
```
mikramarc: dont think this makes sense, if exception is thrown test would just fail anyway. maybe just do:
```self.robot_commander.set_num_planning_attempts(3)```
and add a small comment saying its just testing correctness of syntax?

File: sr_robot_commander/test/test_robot_commander.py [hunk 33]
```diff
@@ -0,0 +1,676 @@
...
+            self.robot_commander.set_num_planning_attempts(3)
+        except (NameError, TypeError):
+            raised = True
+        self.assertFalse(raised)
+
+    def test_get_end_effector_pose_from_state(self):
+        rs = RobotState()
```
mikramarc: robot_state instead of rs

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+
+        for key, value in RA_HOME_ANGLES.items():
```
mikramarc: joint, angle instead of key, value

sr_robot_commander/test/test_robot_commander.py
mikramarc: can you address beppes comments?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+            rs.joint_state.name.append(key)
+            rs.joint_state.position.append(value)
+
+        pose = self.robot_commander.get_end_effector_pose_from_state(rs)
+        self.assertIsInstance(pose, PoseStamped)
+
+    def test_get_planning_frame(self):
+        raised = False
+        try:
```
mikramarc: same as beforee

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+            self.robot_commander.get_planning_frame()
+        except (NameError, TypeError):
+            raised = True
+        self.assertFalse(raised)
+
+    def test_set_and_get_pose_reference_frame(self):
+        reference_frame = "world"
```
mikramarc: const

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+    def test_get_group_name(self):
+        self.assertEqual(self.robot_commander._name, self.robot_commander.get_group_name())
+
+    def test_refresh_named_targets(self):
+        self.robot_commander.refresh_named_targets()
+        condition_1 = type(self.robot_commander._srdf_names) == list
+        condition_2 = type(self.robot_commander._warehouse_names) == list
+        self.assertTrue(condition_1 and condition_2)
+
+    def test_set_max_velocity_scaling_factor__range_ok(self):
```
mikramarc: is double underscore in name deliberate?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+        self.robot_commander.set_max_velocity_scaling_factor(0.2)
```
mikramarc, line 188: add a comment

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+
+    def test_set_max_velocity_scaling_factor__range_not_ok(self):
+        self.assertRaises(MoveItCommanderException, self.robot_commander.set_max_velocity_scaling_factor, 3)
```
mikramarc, line 192: is this a correct syntax...?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+
+    def test_set_max_acceleration_scaling_factor__range_ok(self):
+        self.robot_commander.set_max_acceleration_scaling_factor(0.2)
```
mikramarc, line 196: same

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+
+    def test_set_max_acceleration_scaling_factor__range_not_ok(self):
+        self.assertRaises(MoveItCommanderException, self.robot_commander.set_max_acceleration_scaling_factor, 3)
```
mikramarc, line 200: same

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+
+    def test_plan_to_joint_value_target(self):
+        self.reset_to_home()
+        plan = self.robot_commander.plan_to_joint_value_target(RA_EXAMPLE_TARGET, angle_degrees=False,
+                                                               custom_start_state=None)
+        last_point = list(plan.joint_trajectory.points[-1].positions)
+
+        for i in range(0, len(last_point)):
+            last_point[i] = round(last_point[i], 2)
+            if last_point[i] == -0.00:
```
mikramarc: why do this? also, if you need to change to positive just use abs

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,676 @@
...
+                last_point[i] = 0.00
+        last_point.sort()
```
mikramarc: instead of doing this sorting thigs, why not  just compare two dictionaries?

mikramarc (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
maxzieba: Added

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
maxzieba: The reason is that set_position_target doesn't consider the orientation of the end_effector which is rather not preferred int our case. Agreed with Beatriz to solve it like this. The current solution will move the end effector to the set position with the current orientation.

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
maxzieba: Explained above

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Changed

sr_robot_commander/test/test_robot_commander.py
maxzieba: Removed

sr_robot_commander/test/test_robot_commander.py
maxzieba: Corrected

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Corrected

sr_robot_commander/test/test_robot_commander.py
maxzieba: Changed keys to joint_states. The intersection() part is to get the common keys between 2 joint state - this can be used i.e we would like to compare a joint_state which has 6 joints with another joints_state where there are only 2 joints.

sr_robot_commander/test/test_robot_commander.py
maxzieba: Ok

sr_robot_commander/test/test_robot_commander.py
maxzieba: Ok

sr_robot_commander/test/test_robot_commander.py
maxzieba: Ok

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Ok

sr_robot_commander/test/test_robot_commander.py
maxzieba: Changed

sr_robot_commander/test/test_robot_commander.py
maxzieba: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises
The test passes if the exception is raised. Seems to be correct. 
For the set_max_velocity_scaling_factor, the argument of 3 will raise the exception and the exception is raised (which is expected)

sr_robot_commander/test/test_robot_commander.py
maxzieba: Explained above

sr_robot_commander/test/test_robot_commander.py
maxzieba: Explained above

sr_robot_commander/test/test_robot_commander.py
maxzieba: Added in the top of the file

sr_robot_commander/test/test_robot_commander.py
maxzieba: Yes, only to make the test case a little more visible. But can remove it, no problem

sr_robot_commander/test/test_robot_commander.py
maxzieba: Added

sr_robot_commander/test/test_robot_commander.py
maxzieba: Ok

sr_robot_commander/test/test_robot_commander.py
maxzieba: Added a type check on the pose

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+
+    def compare_poses(self, pose1, pose2, tolerance=0.01):
+        p1_list = [pose1.position.x, pose1.position.y, pose1.position.z,
+                   pose1.orientation.x, pose1.orientation.y, pose1.orientation.z, pose1.orientation.w]
+        p2_list = [pose2.position.x, pose2.position.y, pose2.position.z,
+                   pose2.orientation.x, pose2.orientation.y, pose2.orientation.z, pose2.orientation.w]
+        p1_list = [round(i, 2) for i in p1_list]
+        p2_list = [round(i, 2) for i in p2_list]
+
+        for i in range(0, len(p1_list)):
```
mikramarc: its more "pythonic" to use zip

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+            if abs(p1_list[i]) - abs(p2_list[i]) > tolerance:
```
mikramarc: should be abs(p1 - p2) rather than abs(p1) - abs(p2). This way poses like -5 and 5 and considered equal, which is wrong

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+        cls.eef = cls.robot_commander.get_end_effector_link()
+
+    def reset_to_home(self):
+        self.robot_commander._reset_plan()
+        plan = self.robot_commander.plan_to_joint_value_target(CONST_RA_HOME_ANGLES, angle_degrees=False,
+                                                               custom_start_state=None)
+        self.robot_commander.execute_plan(plan)
+
+    def compare_poses(self, pose1, pose2, tolerance=0.01):
+        p1_list = [pose1.position.x, pose1.position.y, pose1.position.z,
```
mikramarc: change names to pose1_list and pose2_list maybe?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+        p1_list = [round(i, 2) for i in p1_list]
+        p2_list = [round(i, 2) for i in p2_list]
+
+        for i in range(0, len(p1_list)):
+            if abs(p1_list[i]) - abs(p2_list[i]) > tolerance:
+                return False
+        return True
+
+    def compare_joint_states(self, joint_state_1, joint_state_2, tolerance=0.01):
+        common_joint_names = set(joint_state_1.keys()).intersection(set(joint_state_2.keys()))
```
mikramarc: so I get why this is done, but my quesiton is - do we want to consider two joint states that have different set of joints equal if just the shared joints are equal. As in, if the keys are different, doesnt it make the joint_state different in it self? Just asking, not saying.

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+        if len(common_joint_names) == 0:
+            return False
+        for key in common_joint_names:
+            joint_state_1[key] = round(joint_state_1[key], 2)
+            joint_state_2[key] = round(joint_state_2[key], 2)
+            if abs(joint_state_1[key]) - abs(joint_state_2[key]) > tolerance:
```
mikramarc: same abs problem

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+from trajectory_msgs.msg import JointTrajectoryPoint, JointTrajectory
+from std_msgs.msg import Header
+from sensor_msgs.msg import JointState
+from moveit_commander.exception import MoveItCommanderException
+from moveit_commander import conversions
+from actionlib_msgs.msg import GoalStatusArray
+import tf2_ros
+import time
+
+# Some of the test cases are consisting of 1 line. In case of these methods the test verifies if
```
mikramarc:
```suggestion
# Some of the test cases do not have an assert method. In case of these methods the test verifies if
```

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+        pose = self.robot_commander.get_current_pose(reference_frame="world")
+        self.assertTrue(type(pose) == Pose)
+
+    def test_get_current_pose_frame_none(self):
+        pose = self.robot_commander.get_current_pose(reference_frame=None)
+        self.assertTrue(type(pose) == Pose)
+
+    def test_get_current_pose_frame_wrong(self):
+        pose = self.robot_commander.get_current_pose(reference_frame="test_wrong_frame")
+        self.assertTrue(pose is None)
```
mikramarc: use assertIsNone(x)

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+        const_test_names = ["no_target"]
+        output = self.robot_commander.get_named_target_joint_values(const_test_names[0])
+        self.assertTrue(output is None)
+
+    def test_get_end_effector_link(self):
+        self.assertTrue(type(self.robot_commander.get_end_effector_link()) == str)
+
+    def test_get_current_pose_frame(self):
+        pose = self.robot_commander.get_current_pose(reference_frame="world")
+        self.assertTrue(type(pose) == Pose)
```
mikramarc: use assertIsInstance

mikramarc: and in all others before and after

sr_robot_commander/test/test_robot_commander.py
mikramarc: and in all others before and after

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+    def test_get_joints_state(self):
+        ret_val = self.robot_commander.get_joints_state()
+        self.assertTrue(type(ret_val) == JointState)
+
+    def test_run_joint_trajectory(self):
+        self.reset_to_home()
+        trajectory = self.robot_commander.plan_to_joint_value_target(CONST_RA_HOME_ANGLES, angle_degrees=False,
+                                                                     custom_start_state=None).joint_trajectory
+        self.robot_commander.run_joint_trajectory(trajectory)
+        end_state = copy.deepcopy(self.robot_commander.get_current_state())
```
mikramarc: deepcopy tutaj potrzebne?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+        trajectory.append({"joint_angles": {'ra_shoulder_pan_joint': 0.05, 'ra_elbow_joint': 1.95},
+                           "interpolate_time": 0.6, "pause_time": 0.1, "degrees": False})
+        trajectory.append({"joint_angles": {'ra_shoulder_pan_joint': 0.10, 'ra_elbow_joint': 1.90},
+                           "interpolate_time": 0.7, "pause_time": 0.1, "degrees": False})
+        traj = self.robot_commander.make_named_trajectory(trajectory)
+        all_positions = []
+
+        for i in range(0, len(traj.points)):
+            all_positions.append(traj.points[i].positions)
+        for wp in trajectory:
```
mikramarc: whats wp?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -0,0 +1,652 @@
...
+        trajectory = []
+        trajectory.append({"joint_angles": {'ra_shoulder_pan_joint': 0.00, 'ra_elbow_joint': 2.00},
+                           "interpolate_time": 0.5, "pause_time": 0.1, "degrees": False})
+        trajectory.append({"joint_angles": {'ra_shoulder_pan_joint': 0.05, 'ra_elbow_joint': 2.00},
+                           "interpolate_time": 0.4, "pause_time": 0.1, "degrees": False})
+        trajectory.append({"joint_angles": {'ra_shoulder_pan_joint': 0.05, 'ra_elbow_joint': 1.95},
+                           "interpolate_time": 0.6, "pause_time": 0.1, "degrees": False})
+        trajectory.append({"joint_angles": {'ra_shoulder_pan_joint': 0.10, 'ra_elbow_joint': 1.90},
+                           "interpolate_time": 0.7, "pause_time": 0.1, "degrees": False})
+        traj = self.robot_commander.make_named_trajectory(trajectory)
```
mikramarc: "named_trajectory" instead of traj?

mikramarc (changes requested):

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: OK

sr_robot_commander/test/test_robot_commander.py
maxzieba: Ok

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
maxzieba: Sure

sr_robot_commander/test/test_robot_commander.py
maxzieba: waypoint, the same 'naming convention' for it as from sr_robot_commander https://github.com/shadow-robot/sr_interface/blob/noetic-devel/sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py#L575

sr_robot_commander/test/test_robot_commander.py
maxzieba: Okey. The method's name suggests as if the trajectory would be named (i.e. pick_up_box_trajectory) so 'named_trajectory' might be suggestive - tried to avoid it but 'named_trajectory' is better then traj :)

sr_robot_commander/test/test_robot_commander.py
maxzieba: Nope, going to remove

sr_robot_commander/test/test_robot_commander.py
maxzieba: Done

sr_robot_commander/test/test_robot_commander.py
mikramarc: Lol, didnt realise I wrote in polish

sr_robot_commander/test/test_robot_commander.py
maxzieba: Sure

sr_robot_commander/test/test_robot_commander.py
maxzieba: Well, I guess basically the best scenario is to consider both - keys and values to be equal and rename to are_equal_joint_states(), if I'd rename my method to compare_joint_states_by_common_keys() would it make more sense for you?

mikramarc: I guess :P

mikramarc (approved):

sr_multi_moveit/sr_box_ur10_moveit_config/launch/trajectory_execution.launch.xml
```diff
@@ -11,6 +11,7 @@
   <param name="trajectory_execution/allowed_execution_duration_scaling" value="1.2"/> <!-- default 1.2 -->
   <!-- Allow more than the expected execution time before triggering a trajectory cancel (applied after scaling) -->
   <param name="trajectory_execution/allowed_goal_duration_margin" value="0.5"/> <!-- default 0.5 -->
+  <param name="trajectory_execution/allowed_start_tolerance" value="0.1"/> <!-- default 0.01 -->
```
giusebar: if this is needed for the test only I would set it directly there in the code or in the test launch file, so it won't affect the real robot and sim.

giusebar (changes requested):

sr_multi_moveit/sr_box_ur10_moveit_config/launch/trajectory_execution.launch.xml
maxzieba: Added as a param

PR #668, opened 2021-09-09
SRC-6702 Add hybrid controller argument to more launch files

mykolasjuraitis:
## Proposed changes

Argument to use hybrid controller was added to more files.

## Checklist

## Tests

- [x] No tests required to be added. (For small changes that will be tested by CI/CD infrastructure).
- [ ] Added automated tests (if a new class is added (Python or C++), the interface of that class must be unit tested).
- [ ] Manually tested on hardware (if hardware specific or related).

## Documentation

- [x] No documentation required to be added.

### DIFF
--- a/sr_robot_launch/launch/sr_bimanual_ur10arms_hands.launch
+++ b/sr_robot_launch/launch/sr_bimanual_ur10arms_hands.launch
@@ -56,6 +56,8 @@
   <!-- The control mode PWM (true) or torque (false) -->
   <!-- Set to true by default for now as torque control is not available yet -->
   <arg name="pwm_control" default="true"/>
+  <!-- Set to true if you want to use hybrid controller -->
+  <arg name="hybrid_controller" default="false"/>
   <!-- Set the hand type here: hand_e_plus, hand_e, hand_lite, hand_extra_lite -->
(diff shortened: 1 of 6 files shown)


maxzieba (commented):
Do we need to add hybrid controller to launch file which doesn't use a hand? (sr_ur_arm_box)? The hand is here replaced with a box.

mykolasjuraitis:
> Do we need to add hybrid controller to launch file which doesn't use a hand? (sr_ur_arm_box)? The hand is here replaced with a box.

Because https://github.com/shadow-robot/sr_teleop_internal/blob/noetic-devel/sr_teleop_launch/launch/sr_teleop.launch#L127 launches either `sr_ur_arm_box.launch` or `sr_$(arg side)_ur10arm_hand.launch` with the same set of arguments. I guess I could do:
```
<arg if="$(arg hand)" name="hybrid_controller" value="$(arg hybrid_controller)" />
```
Not sure which is best:
1. Have `hybrid_controller` argument everywhere
2. Have conditional arguments where hybrid controller is not used

@mikramarc we need your expert opinion.

biofotis (approved):

maxzieba (approved):

PR #694, opened 2021-12-08
[partial melodic backport] Fixing srdf generation and saving of file

v4hn:
Partial backport of  #615 because https://github.com/ros/xacro/commit/9634128f81d91ead8e92f9b84a3ae2c38890b179 , recently released into ROS melodic, broke the calls
(```object None has no method append```).

@biofotis @beatrizleon

Someone is already working on migrating our PR2 (and thus the mounted hand) to noetic, but it will take more time to do so and it would be great to have the melodic branches still working. :slightly_smiling_face:


biofotis:
i added melodic checks and made them public as well. Could you please re-commit @v4hn ?

v4hn:
> Could you please re-commit @v4hn ?

done

biofotis (approved):

beatrizleon (approved):

PR #716, opened 2022-02-21
fix missing use namespace Eigen

v4hn:
Compiling packages for the [`ros-o` initiative](https://github.com/ros-o/), my build broke because this file implicitly used Eigen datastructures without namespace prefix.

This is a trivial cleanup commit that works just as well with noetic-devel.


v4hn:
CI failure seems unrelated.

beatrizleon:
Thanks for the pr @v4hn. Could you please update your branch?

v4hn:
rebased.

beatrizleon (approved):

PR #727, opened 2022-03-07
B#src 6914 refactor robot commander test

giusebar:
## Proposed changes

Refactoring and cleaning robot commander test.

## Checklist

## Tests

- [ ] No tests required to be added. (For small changes that will be tested by CI/CD infrastructure).
- [ ] Added/Modified automated and PhantomHand CI tests (if a new class is added (Python or C++), the interface of that class must be unit tested).
- [ ] Manually tested on hardware (if hardware specific or related).

## Documentation

- [ ] No documentation required to be added.


sr_robot_commander/test/test_robot_commander.py
```diff
@@ -38,81 +38,128 @@
...
 
 class TestSrRobotCommander(TestCase):
     @classmethod
     def setUpClass(cls):
         rospy.wait_for_message("/move_group/status", GoalStatusArray)
         rospy.wait_for_service("/gazebo/set_model_configuration")
+        rospy.wait_for_message("/clock", Clock)
         rospy.sleep(10.0)  # Wait for Gazebo to sort itself out
         cls.robot_commander = SrRobotCommander("right_arm")
+        cls.robot_commander.set_planner_id("RRTstar")
```
bdenoun: Why using RRTStar? Correct me if I am wrong but we are using BiTRRT, so why not using this one instead in the test?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -38,81 +38,128 @@
...
 TOLERANCE_UNSAFE = 0.04
+PLANNING_ATTEMPTS = 20
 
 
 class TestSrRobotCommander(TestCase):
     @classmethod
     def setUpClass(cls):
         rospy.wait_for_message("/move_group/status", GoalStatusArray)
         rospy.wait_for_service("/gazebo/set_model_configuration")
+        rospy.wait_for_message("/clock", Clock)
```
bdenoun: Don't we want to have a timeout here, just in case one of these topics is not alive for any reason?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -38,81 +38,128 @@
...
+
+        pose_msg.pose.orientation.x = quaternion[0]
+        pose_msg.pose.orientation.y = quaternion[1]
+        pose_msg.pose.orientation.z = quaternion[2]
+        pose_msg.pose.orientation.w = quaternion[3]
+        return pose_msg
+
+    def compare_position(self, initial_position, desired_position, tolerance):
+        initial_position_list = [initial_position.x, initial_position.y, initial_position.z]
+        desired_position_list = [initial_position.x, initial_position.y, initial_position.z]
```
bdenoun: isn't it supposed to be [desired_position.x, desired_position.y, desired_position.z] ? Otherwise regardless of your tolerance you will always have 0

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -38,81 +38,128 @@
...
-        joint_state_1_cpy = copy.deepcopy(joint_state_1)
-        joint_state_2_cpy = copy.deepcopy(joint_state_2)
-        common_joint_names = set(joint_state_1_cpy.keys()).intersection(set(joint_state_2_cpy.keys()))
-        if len(common_joint_names) == 0:
+    def compare_orientation(self, initial_orientation, desired_orientation, tolerance):
+        initial_euler = tf.transformations.euler_from_quaternion([initial_orientation.x, initial_orientation.y,
+                                                                  initial_orientation.z, initial_orientation.w])
+        desired_euler = tf.transformations.euler_from_quaternion([desired_orientation.x, desired_orientation.y,
+                                                                  desired_orientation.z, desired_orientation.w])
+        euler_delta_roll = desired_euler[0] - initial_euler[0]
```
bdenoun: why not directly having abs(desired_euler[0] - initial_euler[0]) instead of adding the abs in the condition below?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -38,81 +38,128 @@
...
+           abs(euler_delta_yaw) > tolerance):
             return False
-        for key in common_joint_names:
-            joint_state_1_cpy[key] = self.normalize_angle_positive(round(joint_state_1_cpy[key], 2))
-            joint_state_2_cpy[key] = self.normalize_angle_positive(round(joint_state_2_cpy[key], 2))
-            if abs(joint_state_1_cpy[key] - joint_state_2_cpy[key]) >= tolerance:
-                return False
         return True
 
+    def compare_poses(self, pose1, pose2, position_threshold=0.02, orientation_threshold=0.04):
```
bdenoun, line 143: That might be what we want, but a tolerated position error of 0.02 is 2cm is I am not mistaken, which seems a lot compared to an orientation error of 0.04 radians (which is 2 degrees)

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -141,9 +186,9 @@ def test_get_group_name(self):
 
     def test_refresh_named_targets(self):
         self.robot_commander.refresh_named_targets()
-        condition_1 = type(self.robot_commander._srdf_names) == list
-        condition_2 = type(self.robot_commander._warehouse_names) == list
-        self.assertTrue(condition_1 and condition_2)
+        is_srdf_list = type(self.robot_commander._srdf_names) == list
+        is_warehouse_names_list = type(self.robot_commander._warehouse_names) == list
```
bdenoun: Don't want to be a pain, but using type(x) == whatever, is usually not the best way to go about it. Slightly better would be type(x) is list, but still not great. Why not using isinstance(x, list) which is about 40% quicker and is usually advised in pep?

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -402,281 +427,182 @@ def test_make_named_trajectory(self):
 
     def test_send_stop_trajectory_unsafe(self):
         self.reset_to_home()
```
bdenoun: Given that you have deleted the previous ones, just making sure you really want to keep this one? Especially that I don't think we need to be in home pose to carry out this test right?

bdenoun (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -488,7 +492,9 @@ def plan_to_named_target(self, name, custom_start_state=None):
         if self.set_named_target(name):
             self.__plan = self._move_group_commander.plan()[CONST_TUPLE_TRAJECTORY_INDEX]
         else:
-            rospy.logwarn("Failed to set to named target")
+            rospy.logwarn("Could not find named target, plan not generated")
+            return False
+        return True
```
maxzieba, line 497: with current method name I'd expect a plan to be returned

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -402,281 +427,182 @@ def test_make_named_trajectory(self):
...
+            is_pose_reached = self.compare_poses(current_pose, desired_pose_msg.pose)
+        self.assertTrue(is_pose_reached)
+
+    def test_move_to_pose_target_unsafe(self):
+        self.reset_to_home()
+        desired_pose_rpy = self.create_test_pose_rpy_from_current_pose()
+        is_pose_reached = False
+        tries = 0
+        desired_pose_msg = self.get_pose_msg_from_pose_rpy(desired_pose_rpy)
+        while not is_pose_reached and tries < 5:
```
maxzieba: maybe make use of PLANNING_ATTEMPTS (same for method below)

sr_robot_commander/test/test_robot_commander.py
```diff
@@ -402,281 +427,182 @@ def test_make_named_trajectory(self):
...
-        rospy.logwarn(after_pose)
-
-        self.assertFalse(condition_1 is True and condition_2 is False)
+        desired_pose_rpy = self.create_test_pose_rpy_from_current_pose()
+        is_pose_reached = False
+        tries = 0
+        desired_pose_msg = self.get_pose_msg_from_pose_rpy(desired_pose_rpy)
+        while not is_pose_reached and tries < 5:
+            tries += 1
+            rospy.loginfo("Moving to pose target attempt {}".format(tries))
```
maxzieba: Do we want the loginfo?

sr_robot_commander/test/test_robot_commander.py
giusebar: will change!

sr_robot_commander/test/test_robot_commander.py
giusebar: wops

sr_robot_commander/test/test_robot_commander.py
giusebar: good point

sr_robot_commander/test/test_robot_commander.py
giusebar: Are we? why? :thinking:

bdenoun: Yes we do: https://github.com/shadow-robot/sr_interface/blob/0f5ee30437627ac81ab7514546762522323d13e1/sr_multi_moveit/sr_multi_moveit_config/config/ur/ompl_planning.yaml#L124

For the why, I would say because the outcome of the iPlanner project was that this one was the best trade-off in many situations

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
giusebar: mmm or just whether it has planned or not. it's not get_plan_ , need to check, it's not great to return a private variable (where we are currently storing the plan)

bdenoun (approved):

maxzieba (approved):

PR #773, opened 2022-10-14
Fixing joints moving to default position when their target is not set

jeferal:
## Proposed changes

DO NOT MERGE YET.

User story: https://shadowrobot.atlassian.net/browse/SP-376

PR to fix the issue desribed in the user story. The issue is that when planning in joint space through the robot commander, for the joints whose target is not specified, a plan would be created from the current state to the default target value, which is 0 or the value in the middle of the range of the joint if 0 is not contained.

This PR uses the set points of the controllers to set the start state of the plan as well as the target state. It needs more testing, unit tests for the hand. In the user story you can see the tests than have been done.

The set points are read through subscriptions to the trajectory controllers for all the joints except from J2 and J1 of the fingers. For these joints, the set point is read from the position controller of J0 and this value is splitted by calculating the underactuation ration from J1 and J2 position states.

If the hand is grasping an object, the joint states might be far from the set points since the object could be not allowing the joint to reach the goal position. In this case, if the difference between the start state of the trajectory (set points) and the state is higher than the allowed_start_tolerance parameter then moveit would throw an error. This is why a function has been added to check if the set point is far from the state and update it with the state if so.

## Checklist

## Tests

- [ ] No tests required to be added. (For small changes that will be tested by CI/CD infrastructure).
- [x] Added/Modified automated and PhantomHand CI tests (if a new class is added (Python or C++), the interface of that class must be unit tested).
- [x] Manually tested on hardware (if hardware specific or related).

## Documentation

- [ ] No documentation required to be added.


sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -15,27 +15,31 @@
 # with this program. If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import
+from __future__ import division
```
maxzieba, line 18: This will be removed with the new lint

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -101,16 +105,63 @@ def __init__(self, name):
 
         controller_list_param = rospy.get_param("/move_group/controller_list")
 
+        robot_description = rospy.get_param("/robot_description")
+        self._joint_limits = dict()
```
maxzieba: better to have this self._joint_limits = {} to be compliant with the new linter (you have also a dictionary init below done with {} style :)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -101,16 +105,63 @@ def __init__(self, name):
...
+        self._set_up_set_points_subscribers(self._controllers)
+
         self.tf_buffer = tf2_ros.Buffer()
         self.listener = tf2_ros.TransformListener(self.tf_buffer)
 
         threading.Thread(None, rospy.spin)
 
+        self._wait_for_set_points()
+
+    def _read_joint_limits(self, robot_description):
```
maxzieba: I think this is supposed to happen only once right? Maybe would be better to rename it to something like _initialize_joint_limits and have then a getter like get_joint_limits (returning self._joint_limits)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -101,16 +105,63 @@ def __init__(self, name):
...
+        robot_dom = minidom.parseString(robot_description)
+        robot = robot_dom.getElementsByTagName('robot')[0]
+
+        for child in robot.childNodes:
+            if child.nodeType is child.TEXT_NODE:
+                continue
+            if child.localName == 'joint':
+                jtype = child.getAttribute('type')
```
maxzieba: maybe joint_type instead jtype? same in 144

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -101,16 +105,63 @@ def __init__(self, name):
...
+                if jtype in ['fixed', 'floating', 'planar']:
+                    continue
+                name = child.getAttribute('name')
+
+                if jtype == 'continuous':
+                    minval = -pi
+                    maxval = pi
+                else:
```
maxzieba, line 155: what is the joint_type passed in this else?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
maxzieba: which seems you have done below :) skip the getter part of the comment :)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +522,75 @@ def get_current_state_bounded(self):
...
+                ratio_part = raw_set_points[joint]/(underactuation_ratio+1)
+                set_point_j1 = underactuation_ratio*ratio_part
+                set_point_j2 = ratio_part
+                set_points.update({f"{joint[:-2]}J1": set_point_j1})
+                set_points.update({f"{joint[:-2]}J2": set_point_j2})
+            elif joint[5:] not in ["J1", "J2"] or joint[3:-2] in ["WR", "TH"]:
+                set_points.update({joint: raw_set_points[joint]})
+
+        joint_state = JointState()
+        joint_state.header = Header()
```
maxzieba: I am not sure if initializing Header is required, correct me if I am wrong :)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -852,6 +1008,47 @@ def _call_action(self, goals):
...
                     goals[client], lambda terminal_state, result: self._action_done_cb(client, terminal_state, result))
 
+    def _set_point_cb(self, msg):
+        """
+        Updates the dictionary mapping joint names with their desired position in the trajectory controllers
+        @param msg - ROS message of type JointTrajectoryControllerState
+        """
+        joint_names_group = self._move_group_commander.get_active_joints()
+        with self._set_points_lock:
+            for i in range(0, len(msg.joint_names)):
```
maxzieba: for i in range(len(msg.joint_names));

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +522,75 @@ def get_current_state_bounded(self):
...
+        within their limits
+        @return - The joint states updated with the joint poisitions bounded
+        """
+        if type(joint_states) == dict:
+            for joint in joint_states:
+                joint_states[joint] = self._bound_joint(joint_states[joint],
+                                                        self._joint_limits[joint])
+
+        elif type(joint_states) == RobotState:
+            for i in range(0, len(joint_states.joint_state.name)):
```
maxzieba: for i in range(len(joint_states.joint_state.name));

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +522,75 @@ def get_current_state_bounded(self):
...
+        return joint_value
+
+    def _bound_state(self, joint_states):
+        """
+        Bounds the joint states within the limits of the joints
+        @param joint_states - It can be of type dict or RobotState. This param should contain the joints to be bounded
+        within their limits
+        @return - The joint states updated with the joint poisitions bounded
+        """
+        if type(joint_states) == dict:
```
maxzieba: if isinstance(joint_states, dict):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -101,16 +105,63 @@ def __init__(self, name):
...
+                        maxval = float(limit.getAttribute('upper'))
+                    except Exception as exception:
+                        rospy.logwarn(f"{exception}: {name} is not fixed, nor continuous, \
+                            but limits are not specified!")
+                        continue
+                self._joint_limits.update({name: (minval, maxval)})
+
+    def get_joint_limits(self):
+        """
+        @return - dictionary mapping joint names with a tuple containing the lower limit
```
maxzieba, line 168: the return type is a dictionary as you define it above, so better to have have it also listed as dictionary in the doc

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +522,75 @@ def get_current_state_bounded(self):
     def get_robot_state_bounded(self):
         return self._move_group_commander._g.get_current_state_bounded()
 
+    @staticmethod
+    def _bound_joint(joint_value, joint_limits):
+        """
+        Forces a joint value to be within the given joint limits
+        @joint_value - Value of the joint to be bounded
+        @joint_limits - Tuple with the lower limit and upper limit of the joint
```
maxzieba, line 553: joint_limits is of type dictionary as defined above

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +522,75 @@ def get_current_state_bounded(self):
...
+        for joint in raw_set_points:
+            if "J0" in joint:
+                # Get j1 j2 ratio from current state
+                underactuation_ratio = current_state[f"{joint[:-2]}J1"]/current_state[f"{joint[:-2]}J2"]
+                ratio_part = raw_set_points[joint]/(underactuation_ratio+1)
+                set_point_j1 = underactuation_ratio*ratio_part
+                set_point_j2 = ratio_part
+                set_points.update({f"{joint[:-2]}J1": set_point_j1})
+                set_points.update({f"{joint[:-2]}J2": set_point_j2})
+            elif joint[5:] not in ["J1", "J2"] or joint[3:-2] in ["WR", "TH"]:
```
maxzieba: Would be good to not use the hardcoded indexes I think, maybe use any() or give a comment on how the joint string is structured?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -770,6 +897,35 @@ def _joint_states_callback(self, joint_state):
...
+                        topics_subscribed.append(topic_name)
+
+        for joint_name in joint_names_group:
+            joint_member = joint_name[3:-2]
+            joint_number = joint_name[5:]
+            if joint_number in ["J1", "J2"] and joint_member not in ["WR", "TH"]:
+                topic_name = f"/sh_{joint_name.lower()[0:5]}j0_position_controller/state"
+                rospy.Subscriber(topic_name,
+                                 JointControllerState,
+                                 self._set_point_j0_cb, f"{joint_name[0:5]}J0",
```
maxzieba: Nice! First time I see using callback_args in use! :smile:

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -770,6 +897,35 @@ def _joint_states_callback(self, joint_state):
...
+            for joint_name in joint_names_group:
+                if joint_name in controllers_list[controller_name]:
+                    topic_name = f"/{controller_name}/state"
+                    if topic_name not in topics_subscribed:
+                        rospy.Subscriber(topic_name, JointTrajectoryControllerState, self._set_point_cb, queue_size=1)
+                        topics_subscribed.append(topic_name)
+
+        for joint_name in joint_names_group:
+            joint_member = joint_name[3:-2]
+            joint_number = joint_name[5:]
```
maxzieba: Add as comment the structure of joint_name, or use any() in the if below if you can :)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: It can be a revolute joint like this one https://github.com/shadow-robot/sr_teleop_hand_internal/blob/2ecab89fbb4fcddd929a07550885bfcdbdc2de83/sr_fingertip_hand_teleop/config/rh_fingers_description.xml#L157

maxzieba: ok, thanks

can you then just move the name = child.getAttribute('name') before the first 'if' to slightly increase readability?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -15,27 +15,31 @@
 # with this program. If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import
```
ethanfowler, line 17: Don't need this anymore

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
ethanfowler: I would consider using a regex to match the joints conditionally

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
ethanfowler: I think @maxzieba is right

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
ethanfowler: Maybe a regex with a substring match?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -852,6 +1012,47 @@ def _call_action(self, goals):
...
                     goals[client], lambda terminal_state, result: self._action_done_cb(client, terminal_state, result))
 
+    def _set_point_cb(self, msg):
+        """
+        Updates the dictionary mapping joint names with their desired position in the trajectory controllers
+        @param msg - ROS message of type JointTrajectoryControllerState
+        """
+        joint_names_group = self._move_group_commander.get_active_joints()
+        with self._set_points_lock:
+            for i in range(len(msg.joint_names)):
```
ethanfowler: Consider using `enumerate()`

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: the attribute name seems to be particular for nodes of type joint

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: Thank you. I refactored this and the tests to check if a joint is underactuated are done with _is_joint_underactuated method that uses regex

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: Thank you, I changed it

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: If I'm not wrong, this repo still has the master [branch of the linter ](https://github.com/shadow-robot/sr_interface/blob/cba8621b259bc947cdfa91f93b27875362ce0264/aws.yml#L12), so if I remove this line, the tests might fail

maxzieba (approved):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -259,9 +318,20 @@ def move_to_joint_value_target(self, joint_states, wait=True,
...
+        # In case that the set points are far from the state, use the state of the joints
+        set_points, robot_state_set_points = self.fix_set_points_if_far_from_state(set_points, robot_state_set_points)
+        # Setting the start state of the plan
+        self._move_group_commander.set_start_state(robot_state_set_points)
+        # Bound the set points to make sure they lay within the joint limits
+        set_points = self._bound_state(set_points)
+        # Set the target of the plan
+        self._move_group_commander.set_joint_value_target(set_points)
         self._move_group_commander.set_joint_value_target(joint_states_cpy)
-        self._move_group_commander.go(wait=wait)
```
jeferal, line 264: It looks like go function uses the current state internally, I tried to find the line of code that does that, but I could not. By performing tests, I could not reproduce the problem on planning when an object is grasped and that makes me think that is using the current state instead of the state set when using .set_start_satate

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -259,9 +318,20 @@ def move_to_joint_value_target(self, joint_states, wait=True,
...
+        set_points, robot_state_set_points = self.get_current_set_points()
+        # In case that the set points are far from the state, use the state of the joints
+        set_points, robot_state_set_points = self.fix_set_points_if_far_from_state(set_points, robot_state_set_points)
+        # Setting the start state of the plan
+        self._move_group_commander.set_start_state(robot_state_set_points)
+        # Bound the set points to make sure they lay within the joint limits
+        set_points = self._bound_state(set_points)
+        # Set the target of the plan
+        self._move_group_commander.set_joint_value_target(set_points)
         self._move_group_commander.set_joint_value_target(joint_states_cpy)
```
beatrizleon, line 331: is it ok to call set_joint_value_target one after the other? or one needs to be deleted?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -95,22 +100,76 @@ def __init__(self, name):
...
         self._controllers = {}
 
+        self._underactuated_joint_finder = re.compile('[r,l]h_[F,M,R,L]FJ[1,2]')
+
         rospy.wait_for_service('compute_ik')
         self._compute_ik = rospy.ServiceProxy('compute_ik', GetPositionIK)
         self._forward_k = rospy.ServiceProxy('compute_fk', GetPositionFK)
 
         controller_list_param = rospy.get_param("/move_group/controller_list")
+        self._allowed_start_tolerance = rospy.get_param("/move_group/trajectory_execution/allowed_start_tolerance", 0.1)
```
toliver, line 110: This module is common for arm and hand (it's the parent of HandCommander and ArmCommander).
How do we distinguish between `allowed_start_tolerance` for arm (likely we want it quite low) and hand?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: The first call of set_joint_value_target sets the set points (or the joint state if the set points are far) as the target for all the joints that belong to the move group. The second call will update only the target of the joints that are passed as an argument (the ones that the user actually wants to move).
This approach is similar to this [recommendation](https://github.com/ros-planning/moveit/issues/3208#issuecomment-1238950199) as well as in the [documentation](http://docs.ros.org/en/jade/api/moveit_ros_planning_interface/html/classmoveit_1_1planning__interface_1_1MoveGroup.html#a8280f121bc31a59de9f07c94e957f1be). 

Another approach would be to first merge the dictionary that contains the set points with the dictionary containing the user targets and then call only once set_joint_value_target.

ethanfowler (approved):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
beatrizleon: got it

beatrizleon (approved):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -770,6 +976,36 @@ def _joint_states_callback(self, joint_state):
...
                                    zip(joint_state.name, joint_state.effort)}
 
+    def _set_up_set_points_subscribers(self, controllers_list):
+        """
+        Sets up the required subscribers to read from the set points of each given controller
+        @param controller_list - Dictionary mapping a trajectory controller with the list of the joints it has
+        """
+        # Get joint names of the group
+        joint_names_group = self._move_group_commander.get_active_joints()
+        trajectory_controllers_subscribed = []
```
toliver: Maybe call it `trajectory_controllers_state_subscribed`

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -770,6 +976,36 @@ def _joint_states_callback(self, joint_state):
...
+        j0_position_controllers_subscribed = []
```
toliver: Maybe `j0_position_controllers_state_subscribed`

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
toliver: Can we add that as a comment?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +545,131 @@ def get_current_state_bounded(self):
...
+                set_point_j0 = raw_set_points[joint_0_name]
+
+                if (state_j1 + state_j2) == 0:
+                    # Avoid division by 0, use directly the set point of J1 and J2
+                    # from the trajectory controller
+                    set_point_j1 = raw_set_points[joint_1_name]
+                    set_point_j2 = raw_set_points[joint_2_name]
+                else:
+                    # Split j0 given state of j1 and j2
+                    set_point_j1 = state_j1 * set_point_j0 / (state_j1 + state_j2)
```
toliver, line 623: Not sure that this is necessary.
If we plan to get a separate set point for J1 and J2, why don't we take the one form the trajectory controller state?

That is the actual one that get set when one uses trajectory messages.

toliver: We should proably rethink if we need to listen to the J0 position controllers at all.

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +545,131 @@ def get_current_state_bounded(self):
...
+
+                set_points.update({joint_1_name: set_point_j1})
+                set_points.update({joint_2_name: set_point_j2})
+
+                # Avoid executing again this
+                underactuation_fingers_processed.append(joint_name[3:5])
+            else:
+                if "J0" not in joint_name:
+                    # Avoind adding set points of J0 to the output
+                    set_points.update({joint_name: raw_set_points[joint_name]})
```
toliver, line 634: Maybe this method only needs to do this bit?

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +545,131 @@ def get_current_state_bounded(self):
...
+            joint_state.position.append(set_points[joint_name])
+
+        move_group_robot_state = RobotState()
+        move_group_robot_state.joint_state = joint_state
+
+        return set_points, move_group_robot_state
+
+    def fix_set_points_if_far_from_state(self, set_points, robot_state):
+        """
+        Updates the set points with the current stateof the robot if the set point
```
toliver: typo stateof

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -464,6 +545,131 @@ def get_current_state_bounded(self):
...
+        @param set_points - Dictionary that maps the joint names with their set points.
+        @param robot_state - RobotState object that contains the joint names and their set points
+        as the position of the joints.
+        @return - Tuple that contains the updated set_points and updated robot_state.
+        """
+        current_state = self.get_current_state()
+
+        # Update set points
+        for joint_name in list(set_points.keys()):
+            if abs(set_points[joint_name] - current_state[joint_name]) > self._allowed_start_tolerance:
```
toliver, line 663: We might want the user to be able to specify the `tolerance` for every use case.
e.g. if you are grasping a big object you might want to allow J3 and J0 set points to be very far from the current positions (set point inside the object while the current pos is on the surface).
Otherwise this "fixing" could result in a dropped object.

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -852,6 +1088,46 @@ def _call_action(self, goals):
...
+                        if self._is_joint_underactuated(joint_name) and f"{joint_name[0:5]}J0" not in joint_names_group:
+                            joint_names_group.append(f"{joint_name[0:5]}J0")
+                    for joint_name in joint_names_group:
+                        if joint_name not in self._set_points.keys():
+                            break
+                    else:
+                        self._are_set_points_ready = True
+                        self._set_points_cv.notifyAll()
+
+    def _set_point_j0_cb(self, msg, joint_name):
```
toliver, line 1114: As mentioned before, not really sure if we need to know the J0 goals at all.
Given that we know the J1 and J2 goals (because this whole RobotCommander relies on having a JointTrajectoryposition controller running), maybe we don't need this?

toliver (changes requested):

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
```diff
@@ -770,6 +976,36 @@ def _joint_states_callback(self, joint_state):
             self._joints_effort = {n: v for n, v in
                                    zip(joint_state.name, joint_state.effort)}
```
maxzieba, line 977: side note -> this can be just self._joints_effort = {zip(joint_state.name, joint_state.effort)}

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: In there I am only reading that parameter. When launching `roslaunch sr_robot_launch sr_ur_arm_hand.launch` it seems (at least in simulation) that `allowed_start_tolerance` is set in [here](https://github.com/shadow-robot/sr_interface/blob/14b5e9f4eed0cb707b0f922e90a63bea6c2a3715/sr_multi_moveit/sr_multi_moveit_config/launch/trajectory_execution.launch.xml#L16). And it looks like there is only that parameter for any group:
image
Correct me if I am wrong please.

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: Changed in [12e6217](https://github.com/shadow-robot/sr_interface/pull/773/commits/12e6217962cf52a1bd7a5781fe1b183cd3da3797)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: Changed in [12e6217](https://github.com/shadow-robot/sr_interface/pull/773/commits/12e6217962cf52a1bd7a5781fe1b183cd3da3797)

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: I think we took the decision of listening to J0's set points because the J1 and J2 set points because of the underactuation might be far from the real state and also outside of the joint limits. For instance, when the set point of FFJ2 is 90deg and the set point of FFJ1 is ~60deg I find this:
image
If I directly set these set points as the target state when planning, moveit will throw the error: target state is outside of the bounds. I think this also depends on a tolerance that is set through this [method](http://docs.ros.org/en/jade/api/moveit_commander/html/classmoveit__commander_1_1move__group_1_1MoveGroupCommander.html#a77fb81f3af386a0faab7554af48a1476).
However, since we also encountered the "problem" of the set points being far from the state when grasping an object and added the "fix" of checking if the set point is far from the state and if it is, use the state then listening to j0's might not be useful.
I also agree that perhaps we should rethink of this whole architecture again.

sr_robot_commander/src/sr_robot_commander/sr_robot_commander.py
jeferal: changed in [543eb52](https://github.com/shadow-robot/sr_interface/pull/773/commits/543eb5221b70e3838e7b482b0f64ea9403860b41)

